PR #5, opened 2019-01-28
Adding a class for reading data format of Glory Duck project.

dkerszberg:


include/GloryDuckTables2019Lkl.h
```diff
@@ -0,0 +1,74 @@
...
+{
+ public:
+  // constructors
+  GloryDuckTables2019Lkl(TString inputString="");
+  
+  // destructor
+  virtual ~GloryDuckTables2019Lkl();
+ 
+  // make the TObjArray own its contents
+  void SetOwner(Bool_t set=kTRUE)            {fSampleArray->SetOwner(set);}
```
javierrico: Why is this needed? it is not in the Fermi version.
In any case either setting ownership is needed or not, but should not be a decision of the user, should it?
I mean, if it is needed, just hard-code the call in the appropriate place, right?

dkerszberg: It is not needed (at least I don't use it). I added it when I created the TObjectArray for the class (inspired by the JointLkl class) and at that time I copied all available function concerning the TObjectArray just in case I needed them later... I will delete that line.

include/GloryDuckTables2019Lkl.h
```diff
@@ -0,0 +1,74 @@
...
+
+  // add samples
+  inline void AddSample(Lkl* sample)        {fSampleArray->Add(sample); SetChecked(kFALSE);}
+
+  // getters
+  inline const TObjArray* GetSampleArray()     const {return fSampleArray;}
+  inline       Int_t      GetNSamples()        const {return fSampleArray->GetEntries();}
+  inline       Lkl*       GetReferenceSample() const {return (Lkl*)fSampleArray->At(0);}
```
javierrico: This line you don't need. In JointLkl the reference sample is the one for which units are the same as in the JointLkl, and all the rest are related to the reference by the units. Here there is no need since the samples are not part of a joint Lkl, but different alternatives depending on the value of the mass, and by construction they all have the same units

include/GloryDuckTables2019Lkl.h
javierrico: Ok, I need to revise the use of JointLkl::SetOwner, because my comment applies also there

include/GloryDuckTables2019Lkl.h
```diff
@@ -0,0 +1,74 @@
...
+  inline       Lkl*       GetSample(Int_t i)   const {return (Lkl*)fSampleArray->At(i);}
+  inline       Double_t   GetLogJ()            const {return fLogJ;}
```
javierrico: This and fLogJ are not needed, since the values of <sv> are for some assumed value of J, that will need to be provided externally.
I'm actually wondering if we should not propose to the GD group that the data we change is lkl vs <sv>*J, so that no value of J needs to be assumed, which would make the thing less prone to mistakes (e.g assuming different J values for same object by different instruments)

include/GloryDuckTables2019Lkl.h
```diff
@@ -0,0 +1,74 @@
...
+  inline       UInt_t     GetNMasses()         const {return fNMasses;}
+  inline       UInt_t     GetActiveMass()      const {return fActiveMass;}
+
+  // unable SetUnitsOfG, dPhi/dE_signal must have proper units
+  virtual void SetUnitsOfG(Double_t unit);
```
javierrico: On a second thought, the SetUnitsOfG should not be deactivated, because who knows if needed to convert e.g. to <sv>*J or <sv> in different units. I will also change that in the FermiTables class

include/GloryDuckTables2019Lkl.h
```diff
@@ -0,0 +1,74 @@
...
+
+  // set mass index to the current active one
+  void SetActiveMass(Double_t mass);
```
javierrico: This function should return an Int_t (0 in case of success, 1 otherwise -this is actually a general rule for functions that don't need to return anything and that can fail depending on input etc)

include/GloryDuckTables2019Lkl.h
```diff
@@ -0,0 +1,74 @@
...
+  inline  void  ClearSampleList()  {fSampleArray->Clear();}
+
+  // data input
+  Int_t CreateAndAddNewParabola(Double_t mass,Double_t svmin,Double_t svmax,Int_t npoints,Double_t* sv,Double_t* logL);
+
+ private:
+  UInt_t      fNMasses;       //   Number of masses
+  UInt_t      fActiveMass;    //   Position in the array fMass of the current active mass
+  TObjArray*  fSampleArray;   //-> Array of Lkl objects (one for each mass) 
+  Double_t*   fsvMin;         //-> [fNMasses] array with <sv> lower values
```
javierrico: I think we need a list of sv values, which is valid for all masses, right? that reflects the format we decided, doesn't it?
so not this line nor the following but rather 
Double_t* fsvVals

javierrico (changes requested):
I guess you should see now several comments at once, I'm not finished yet, but I am for today, so I send you this bunch and I will try to finish tomorrow

src/GloryDuckTables2019Lkl.cc
```diff
@@ -0,0 +1,403 @@
...
+  Lkl* sample;
+  while((sample=(Lkl*)iter->Next()))
+    {
+      sample->InitMinuit();
+      sample->SetMinuitLink();
+    }
+  delete iter;
+
+  Double_t pStart[gNPars] = {(ginit==0? gRefsv :ginit)};
+  Double_t pDelta[gNPars] = {1};
```
javierrico: 1 is not the natural step for <sv>, probably something like gRefsv would be better.
Needs to be tested with realistic file

src/GloryDuckTables2019Lkl.cc
```diff
@@ -0,0 +1,403 @@
...
+  SetParameters(gParName,pStart,pDelta);
+}  
+
+////////////////////////////////////////////////////////////////
+//
+// Check that everything is correct
+//
+// Return 0 in case of success, 1 otherwise
+//
+Int_t GloryDuckTables2019Lkl::MakeChecks()
```
javierrico, line 153: Add check that the different data members (fNMasses, fsv,fMass,...) exist and have sensible values (e.g. not NULL)

src/GloryDuckTables2019Lkl.cc
```diff
@@ -0,0 +1,403 @@
...
+
+  cout << "GloryDuckTables2019Lkl::ReadGloryDuckInputData (" << GetName() << ") Message: Read " <<  GetSampleArray()->GetEntries() << " parabolas for a total of " << fNMasses << " DM masses." << endl;
+  return 0;
+}
+
+////////////////////////////////////////////////////////////////
+//
+// Add a new ParabolaLkl object representing the -2logL values vs
+// <sv> for one DM mass
+//
```
javierrico, line 274: Describe each of the arguments of this function

include/GloryDuckTables2019Lkl.h
javierrico: While revising the .cc file, I realize that the list of <sv> values is stored in every ParabolaLkl object. Then I think you do not need to keep it also here, you can always access the list of <sv> values accessing any of the parabolas.
Maybe you want to add a function to the GloryDuckTables2019 class that acts as a shortcut to access the list of <sv> values from one of the Parabolas objects (e.g. the first one).

This assumes that all parabolas have the same <sv> list, which is enforced at the definition of the format only, but not in the implementation of the class. Maybe at the level of implementation we don't want to assume it, to be more flexible in view of possible changes of format. 

Al alternative to force that all <sv> values for all Parabolas are the same is to store the list of <sv> only once at the GloryDuckTables2019 level and in the Parabola terms, in the x axis instead of <sv> you just store consecutive integers from 0 to nSvVals-1

src/GloryDuckTables2019Lkl.cc
```diff
@@ -0,0 +1,403 @@
...
+// Return 0 in case of success, 1 otherwise
+//
+Int_t GloryDuckTables2019Lkl::CreateAndAddNewParabola(Double_t mass,Double_t svmin,Double_t svmax,Int_t npoints,Double_t* sv,Double_t* logL)
```
javierrico: Definitely, svmin and svmax are not needed, they can always be read from sv, which is stored in every parabola

src/GloryDuckTables2019Lkl.cc
```diff
@@ -0,0 +1,403 @@
...
+  SetChecked(kFALSE);
+  return 0;
+}
+
+//////////////////////////////////////////////////////////////////
+//
+// Function called by Lkl::PrintOverview
+// Print info about the experimental data
+//
+void GloryDuckTables2019Lkl::PrintData(Int_t level)
```
javierrico, line 310: Print also all parabolas contents. Maybe later can be removed if we get too messy logs, but for test purposes it will be useful

src/GloryDuckTables2019Lkl.cc
```diff
@@ -0,0 +1,403 @@
...
+// 
+// Return a pointer to the canvas, in case you want to modify it
+//
+TCanvas* GloryDuckTables2019Lkl::PlotInputData() 
+{
+  const Int_t nlines = 5;  // number of lines in parabolas canvas
+  Int_t ncols = TMath::Ceil(fNMasses/Float_t(nlines)); // number of columns in parabolas canvas
+
+  // create and divide canvas
+  TCanvas* canvas = new TCanvas("inputDataCanvas","-2logL value per energy bin and energy flux",ncols*250,nlines*250);
```
javierrico: Fix title of the canvas

src/GloryDuckTables2019Lkl.cc
```diff
@@ -0,0 +1,403 @@
...
+  return canvas;
+}
+
+////////////////////////////////////////////////////////////////////////
+//
+// Set mass index to the current active one
+//
+void GloryDuckTables2019Lkl::SetActiveMass(Double_t mass)
+{
+  for (Int_t counter = 0; counter < fNMasses; counter++)
```
javierrico, line 393: Add comment briefly describing what you are checking

src/GloryDuckTables2019Lkl.cc
```diff
@@ -0,0 +1,403 @@
...
+      gPad->Update();
+    }
+  return canvas;
+}
+
+////////////////////////////////////////////////////////////////////////
+//
+// Set mass index to the current active one
+//
+void GloryDuckTables2019Lkl::SetActiveMass(Double_t mass)
```
javierrico: We need another function that can have the same name, but takes some Int_t index as input, and sets the active mass to fMass[index]. Only check needed is that index is non negative and  <fNMasses

src/GloryDuckTables2019Lkl.cc
```diff
@@ -0,0 +1,403 @@
...
+  f = 0;
+
+  // get internal object
+  GloryDuckTables2019Lkl* mylkl = dynamic_cast<GloryDuckTables2019Lkl*>(minuit->GetObjectFit());
+
+  // g is the global scale for dPhidESignal
+  Double_t g = par[0];
+
+  // get likelihood for current active mass
+  Lkl* sample = mylkl->GetSample(mylkl->GetActiveMass());
```
javierrico: This won't work, because sample will be interpreted as a Lkl object, not as ParabolaLkl one, use the dynamic cast as in the FermiTables2016Lkl case: 
ParabolaLkl* sample = dynamic_cast<ParabolaLkl*>mylkl->GetSample(mylkl->GetActiveMass());

javierrico (changes requested):
I think this completes my review. Now I wait for you to address all comments before I have another look.

dkerszberg:
I committed all the changes according to your review except for the one concerning SetUnitsOfG() because so far we didn't change the data format.

include/GloryDuckTables2019Lkl.h
javierrico: I see you did not do this because "so far we didn't change the data format", which I think is not a good reason. Even for a data format consisting on <sv> values the units are not univoque, so in that sense I think the active version of SetUnitsOfG is needed

include/GloryDuckTables2019Lkl.h
javierrico: I think that you do not need the fsvVals either, because those values are stored in every Parabola, is there any need to have them again saved as data members of GloryDuck class? Probably it's enough to add a function GetSvVals that access them from one of the Parabolas (e.g. the first one).

javierrico:
A couple more of details and we're ready to go...

javierrico (approved):

javierrico:
Pulled

PR #9, opened 2019-03-28
New Gloryduck format

TjarkMiener:
In this PR, I modify the Gloryduck class to read the new Gloryduck format and add ability to ignore the J-Factor at the beginning of the file.

### DIFF
--- a/include/GloryDuckTables2019Lkl.h
+++ b/include/GloryDuckTables2019Lkl.h
@@ -58,6 +58,7 @@ class GloryDuckTables2019Lkl : public Lkl
   Double_t*   fMass;          //-> [GeV] array with dark matter particle mass
   TObjArray*  fSampleArray;   //-> Array of Lkl objects (one for each mass) 
   UInt_t      fNsvVals;       //   Number of <sv> values
+  Double_t    logJ;           //   Value of logJ
 
   ClassDef(GloryDuckTables2019Lkl,1) // Likelihood vs <sv>
 };
--- a/src/GloryDuckTables2019Lkl.cc
+++ b/src/GloryDuckTables2019Lkl.cc
@@ -207,51 +207,73 @@ Int_t GloryDuckTables2019Lkl::ReadGloryDuckInputData(TString filename)
   else
     cout << "GloryDuckTables2019Lkl::ReadGloryDuckInputData (" << GetName() << ") Message: Reading masses and <sigmav> values from file " << filename << endl;
 
-  vector<Double_t> vsigma ;    // vector with <sv> values
-  vector<Double_t> vlkl ;      // vecot with -2logL values
+  // get the number of lines of the file
+  Int_t numRows = 0;
+  string line;
+  while (getline(ff, line))
+    numRows++;
+  ff.clear();
+  ff.seekg(0, ios::beg);
 
-  // get first line of the file (with <sv> values)
-  string sigmavLine;
-  getline(ff,sigmavLine);
+  vector<Double_t> sigmav ;                   // vector with <sv> values
+  vector<Double_t> mass ;                     // vector with mass values
+  vector<vector<Double_t> > vlkl2D(numRows-1);// 2D vector with -2logL values
+  vector<Double_t> vlkl;                      // vector with -2logL values
+
+  // get first line of the file (with mass values)
+  string massLine;
+  getline(ff,massLine);
 
   // read first line of the file
-  std::istringstream row(sigmavLine);
+  std::istringstream first_line(massLine);
   Double_t field;
-  while (row>> field)
+  while (first_line >> field)
     {
+      if(fNsvVals != 0)
+        mass.push_back(field);
       fNsvVals++;
-      vsigma.push_back(field);
     }
 
-  Double_t readingMass = 0.;
+  Double_t readingSigmav = 0.;
   Double_t readingLkl = 0.;
-  Int_t counter = 0;
+  Int_t col = 0;
+  Int_t row = 0;
 
-  // get mass (first column)
-  while(ff >> readingMass)
+  // get <sv> (first column)
+  while(ff >> readingSigmav)
     {
-      // get rest of the line (<sv> values)
+      // store the <sv> values
+      sigmav.push_back(readingSigmav);
+
+      // get rest of the line
       string LklLine;
       getline(ff,LklLine);
 
       // read rest of the line
+      vlkl2D[col] = vector<Double_t>(mass.size());
+      row = 0;
       std::istringstream line(LklLine);
       while (line >> readingLkl)
         {
-          vlkl.push_back(readingLkl);
+          vlkl2D[col][row] = readingLkl;
+          row++;
         }
-
-      // add new parabola
-      CreateAndAddNewParabola(readingMass,vsigma.size(),vsigma.data(),vlkl.data());
-
-      // check if line was complete or not
-      if(vlkl.size() != vsigma.size())
+      // check if line was complete or not HERE -1 LATER
+      if(vlkl2D[col].size() != mass.size())
         {
-          cout << "You have a different line size length for line " << counter+1 << " ! Number of sigmav values = " << vsigma.size() << " while number of likelihood values = " << vlkl.size() << endl; 
+          cout << "You have a different line size length for line " << col+1 << " ! Number of mass values = " << mass.size() << " while number of likelihood values = " << vlkl2D[col].size() << endl;
         }
-      counter++;
-      vlkl.clear();
-   }
+      col++;
+    }
+
+  // add new parabolas
+  for (row = 0; row < fNsvVals-1; row++) {
+    for (col = 0; col < numRows-1; col++) {
+      vlkl.push_back(vlkl2D[col][row]);
+    }
+    CreateAndAddNewParabola(mass[row],sigmav.size(),sigmav.data(),vlkl.data());
+    vlkl.clear();
+  }
 
   ff.close(); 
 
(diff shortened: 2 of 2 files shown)


src/GloryDuckTables2019Lkl.cc
```diff
@@ -207,51 +207,73 @@ Int_t GloryDuckTables2019Lkl::ReadGloryDuckInputData(TString filename)
...
-  vector<Double_t> vlkl ;      // vecot with -2logL values
-
-  // get first line of the file (with <sv> values)
-  string sigmavLine;
-  getline(ff,sigmavLine);
+  // get the number of lines of the file
+  Int_t numRows = 0;
+  string line;
+  while (getline(ff, line))
+    numRows++;
```
dkerszberg: Here you can simply increment fNsvVals directly. Just remember to subtract 1 because there is one more line than sv values.

src/GloryDuckTables2019Lkl.cc
```diff
@@ -207,51 +207,73 @@ Int_t GloryDuckTables2019Lkl::ReadGloryDuckInputData(TString filename)
...
     cout << "GloryDuckTables2019Lkl::ReadGloryDuckInputData (" << GetName() << ") Message: Reading masses and <sigmav> values from file " << filename << endl;
 
-  vector<Double_t> vsigma ;    // vector with <sv> values
-  vector<Double_t> vlkl ;      // vecot with -2logL values
-
-  // get first line of the file (with <sv> values)
-  string sigmavLine;
-  getline(ff,sigmavLine);
+  // get the number of lines of the file
+  Int_t numRows = 0;
```
dkerszberg: You can remove this line. See comment 3 lines below.

src/GloryDuckTables2019Lkl.cc
```diff
@@ -207,51 +207,73 @@ Int_t GloryDuckTables2019Lkl::ReadGloryDuckInputData(TString filename)
...
+  string line;
+  while (getline(ff, line))
+    numRows++;
+  ff.clear();
+  ff.seekg(0, ios::beg);
+
+  vector<Double_t> sigmav ;                   // vector with <sv> values
+  vector<Double_t> mass ;                     // vector with mass values
+  vector<vector<Double_t> > vlkl2D(numRows-1);// 2D vector with -2logL values
```
dkerszberg: You can replace numRows-1 by fNsvVals. See comment above.

src/GloryDuckTables2019Lkl.cc
```diff
@@ -207,51 +207,73 @@ Int_t GloryDuckTables2019Lkl::ReadGloryDuckInputData(TString filename)
...
   // read first line of the file
-  std::istringstream row(sigmavLine);
+  std::istringstream first_line(massLine);
   Double_t field;
-  while (row>> field)
+  while (first_line >> field)
     {
+      if(fNsvVals != 0)
+        mass.push_back(field);
       fNsvVals++;
```
dkerszberg: Here it should be the number of masses fNMasses that increases not fNsvVals. But it is already increased at the first line of CreateAndAddNewParabola which I think is the right place for it so I would just remove it.

src/GloryDuckTables2019Lkl.cc
```diff
@@ -207,51 +207,73 @@ Int_t GloryDuckTables2019Lkl::ReadGloryDuckInputData(TString filename)
...
+  getline(ff,massLine);
 
   // read first line of the file
-  std::istringstream row(sigmavLine);
+  std::istringstream first_line(massLine);
   Double_t field;
-  while (row>> field)
+  while (first_line >> field)
     {
+      if(fNsvVals != 0)
```
dkerszberg: Rather than skipping the first cell (logJ) you should store it in a private variable of the class so that later we can check that all logJ values from a same source are the same.

src/GloryDuckTables2019Lkl.cc
```diff
@@ -207,51 +207,73 @@ Int_t GloryDuckTables2019Lkl::ReadGloryDuckInputData(TString filename)
...
-      // get rest of the line (<sv> values)
+      // store the <sv> values
+      sigmav.push_back(readingSigmav);
+
+      // get rest of the line
       string LklLine;
       getline(ff,LklLine);
 
       // read rest of the line
+      vlkl2D[col] = vector<Double_t>(mass.size());
```
dkerszberg: From here until the end I think you exchange row and col. For instance line 259 it is the number of col that should be incremented while you are reading a line/row.

src/GloryDuckTables2019Lkl.cc
```diff
@@ -207,51 +207,73 @@ Int_t GloryDuckTables2019Lkl::ReadGloryDuckInputData(TString filename)
...
+          row++;
         }
-
-      // add new parabola
-      CreateAndAddNewParabola(readingMass,vsigma.size(),vsigma.data(),vlkl.data());
-
-      // check if line was complete or not
-      if(vlkl.size() != vsigma.size())
+      // check if line was complete or not HERE -1 LATER
+      if(vlkl2D[col].size() != mass.size())
```
dkerszberg: I tried to remove up to 3 values from my test file but I never got the warning that I had a shorter line. I think it is because line 253 you declare the vector already with a size same as the mass vector but I didn't had time to investigate more.

src/GloryDuckTables2019Lkl.cc
```diff
@@ -207,51 +207,73 @@ Int_t GloryDuckTables2019Lkl::ReadGloryDuckInputData(TString filename)
...
+          cout << "You have a different line size length for line " << col+1 << " ! Number of mass values = " << mass.size() << " while number of likelihood values = " << vlkl2D[col].size() << endl;
         }
-      counter++;
-      vlkl.clear();
-   }
+      col++;
+    }
+
+  // add new parabolas
+  for (row = 0; row < fNsvVals-1; row++) {
```
dkerszberg: I'm a bit lost here because of my own comment above but I believe (following the logic of the file) it should be something like
for (col = 0; col < fNMasses; col++) {
  for (row = 0; row < fNsvVals; row++) {
so that you create as many parabolas as there is masses and the parabolas are of length the number of sv values.

dkerszberg (commented):
Hi Tjark,
Here are my comments.
Cheers,
Daniel

src/GloryDuckTables2019Lkl.cc
tjarkmiener: Okay! I think increasing in the first line of CreateAndAddNewParabola is the better way!

src/GloryDuckTables2019Lkl.cc
tjarkmiener: Could be! I will check and change it!

src/GloryDuckTables2019Lkl.cc
TjarkMiener: This could be solved while reading the file like mentioned above!

src/GloryDuckTables2019Lkl.cc
tjarkmiener: I think the code is doing what you mention here, but the names are wrong!

TjarkMiener:
Hi @dkerszberg! Thanks for the useful comments! I adopted all of them and tested the code! It's working stable and raise a error message when insert an inappropriate input!

src/GloryDuckTables2019Lkl.cc
```diff
@@ -207,17 +207,17 @@ Int_t GloryDuckTables2019Lkl::ReadGloryDuckInputData(TString filename)
   else
     cout << "GloryDuckTables2019Lkl::ReadGloryDuckInputData (" << GetName() << ") Message: Reading masses and <sigmav> values from file " << filename << endl;
 
-  // get the number of lines of the file
-  Int_t numRows = 0;
+  // get the number of lines (=number of <sv> values - 1) in the file
```
dkerszberg: To be more clear I would write something like:

// get the number of <sv> values (which is one less than the number of lines in the file)

src/GloryDuckTables2019Lkl.cc
```diff
@@ -250,28 +255,27 @@ Int_t GloryDuckTables2019Lkl::ReadGloryDuckInputData(TString filename)
...
-          row++;
+          vlkl2D[row][col] = readingLkl;
+          col++;
         }
-      // check if line was complete or not HERE -1 LATER
-      if(vlkl2D[col].size() != mass.size())
+      if(col != mass.size())
         {
-          cout << "You have a different line size length for line " << col+1 << " ! Number of mass values = " << mass.size() << " while number of likelihood values = " << vlkl2D[col].size() << endl;
+          cout << "You have a different line size length for line " << row+1 << " ! Number of mass values = " << mass.size() << " while number of likelihood values = " << col << endl;
```
dkerszberg: Here you nedd "row+2" instead of "row+1" because of the first line that contains logJ and masses.

dkerszberg (commented):
I have only 2 (very) smalls comments. Appart from that it all works fine :-)

src/GloryDuckTables2019Lkl.cc
tjarkmiener: Okay, I will change this! I started counting from 0 ;)

src/GloryDuckTables2019Lkl.cc
tjarkmiener: Ok!

include/GloryDuckTables2019Lkl.h
```diff
@@ -58,6 +58,7 @@ class GloryDuckTables2019Lkl : public Lkl
   Double_t*   fMass;          //-> [GeV] array with dark matter particle mass
   TObjArray*  fSampleArray;   //-> Array of Lkl objects (one for each mass) 
   UInt_t      fNsvVals;       //   Number of <sv> values
+  Double_t    logJ;           //   Value of logJ
```
javierrico, line 61: please use fLogJ to follow coding convention

javierrico (approved):

javierrico:
Hi, I have merged your PR, just consider the minor, coding convention comment for your next PR

Issue #7, opened 2019-02-25
Include possibility of creating arbitrary linear combination of dN/dE histograms


javierrico:
scripts/jointLklDM.C reads dN/dE histograms from existing files. So far, histograms provided by Cirelli et al. for 100% branching ratios into selected SM pairs are included in gLike distribution. 
Simple modification of scripts/jointLklDM.C would allow reading branching ratio values for different channels from the rc file, in order to produce arbitrary linear combinations of existing dN/dE histograms

javierrico:
Done by Tjark (PR #11 )

PR #11, opened 2019-04-09
Branon dark matter model #7

TjarkMiener:
This is a preliminary PR for Issue #7. You can test the code with execute the ROOT Macro testBranonDMSearches.C (`.x testBranonDMSearches.C+`). You can change BRs and annihilation channels in the script! So far for testing you have to do it by hand. Once, everything is working and have been tested, I will modify the rcfiles and the scripts!


src/Iact1dUnbinnedLkl.cc
```diff
@@ -889,64 +889,78 @@ Int_t Iact1dUnbinnedLkl::ReaddNdESignal(TString filename)
...
+  TFile* dNdESignalFile[nFiles];
+  TH1F*  hProvdNdESignal[nFiles];
   TH1F*  linCHdNdESignal;
 
   Int_t status      = 0;
   Int_t nBins       = 0;
   Double_t lowEdge  = 0.0;
   Double_t highEdge = 0.0;
-  for(Int_t file=0;file<numFiles;file++)
+  // open file and look for histograms
```
javierrico: In general, please a bit more of comments, e.g. here you should add "check that all histograms have same binning"

src/Iact1dUnbinnedLkl.cc
```diff
@@ -889,64 +889,78 @@ Int_t Iact1dUnbinnedLkl::ReaddNdESignal(TString filename)
...
-      if(!hProvdNdESignal[file]) status=1;
+      dNdESignalFile[iFile]  = new TFile(filenames[iFile]);
+      hProvdNdESignal[iFile] = (TH1F*) dNdESignalFile[iFile]->Get("hdNdE");
+      if(!hProvdNdESignal[iFile]) status=1;
       if(nBins!=0)
         {
-          if(hProvdNdESignal[file]->GetNbinsX()!=nBins) status=1;
-          if(hProvdNdESignal[file]->GetBinLowEdge(1)!=lowEdge) status=1;
-          if(hProvdNdESignal[file]->GetBinLowEdge(nBins+1)!=highEdge) status=1;
+          if(hProvdNdESignal[iFile]->GetNbinsX()!=nBins) status=1;
```
javierrico: This forces you to have all dNdE histograms with same binning, which may be true for the Cirelli distribution, but probably we don't want it to be a strict rule. I suggest you use the internal funcion readAndInterpolate to make sure all histograms have the same binning before you sum them

src/Iact1dUnbinnedLkl.cc
```diff
@@ -889,64 +889,78 @@ Int_t Iact1dUnbinnedLkl::ReaddNdESignal(TString filename)
...
         }
      }
 
   // perform the linear combination
   if(status==0)
     {
-      Double_t linC;
-      for(Int_t ibin=1;ibin<nBins+1;ibin++)
+      Double_t linC = 0;
+      for(Int_t ibin=0;ibin<nBins;ibin++)
```
javierrico: I strongly believe this is not correct. You are re-creating linCHdNdESignal for each bin in the original dNdE histograms. linCHdNdESignal should be created just once at the beginning with the same binning as the (probably modified by the readAndInterpolate function as suggested in precious comment) original dNdE histograms, and then all of them should be added by nFiles calls to TH1::Add (please read some subtleties of this function in the documentation: https://root.cern.ch/root/html526/TH1.html#TH1:Add%1). No loop on nBins is needed

javierrico (changes requested):
I don't think the code can make anything sensible without applying the changes suggested in the last comment of this review, so I stop here and wait for the next commit to resume revision

src/Iact1dUnbinnedLkl.cc
tjarkmiener: Yes, that's true! I should have spotted that stupid mistake! The new commit fixed the bug and used TH1::Add for the linear combination of the histograms. This bug caused the previous problems, because I overwrote the bin content every time. Now the code runs smoothly and give reasonable results! Setting one annihilation channel to 100% is giving us the same results as with conventional function.

src/Iact1dUnbinnedLkl.cc
```diff
@@ -932,26 +932,17 @@ Int_t Iact1dUnbinnedLkl::ReaddNdESignal(Int_t nFiles, TString* filenames, Float_
         }
      }
 
-  // perform the linear combination
   if(status==0)
     {
-      Double_t linC = 0;
-      for(Int_t ibin=0;ibin<nBins;ibin++)
+      // create histogram for the linear combination of dNdESignal
+      linCHdNdESignal = new TH1F("linCHdNdESignal","dN/dE for signal events",nBins,lowEdge,highEdge);
```
javierrico: No need to define title, or axis titles, because this is just a provisional histogram and that information will never be used.

javierrico (changes requested):
Please consider the other comments of the previous review

src/Iact1dUnbinnedLkl.cc
tjarkmiener: Done!

src/Iact1dUnbinnedLkl.cc
tjarkmiener: Good point! Solved in the new commit!

src/Iact1dUnbinnedLkl.cc
```diff
@@ -889,6 +889,78 @@ Int_t Iact1dUnbinnedLkl::ReaddNdESignal(TString filename)
...
+// Then perform a linear combination
+// dN/dE = sum(BR_i * dN_i/dE).
+// Replacement of existing file is allowed
+// Return 0 in case of success
+//        1 if file is not found
+//
+Int_t Iact1dUnbinnedLkl::ReaddNdESignal(Int_t nFiles, TString* filenames, Float_t* branchingRatios)
+{
+  // Initialize the array of pointers and other variables
+  TFile* dNdESignalFile[nFiles];
```
javierrico: I don't think you need an array of pointer for dNdESignalFile or hProvdNdESignal, since they are used only inside the loop over nFiles, therefore you can have just one pointer and reuse it at each step of the loop.

src/Iact1dUnbinnedLkl.cc
```diff
@@ -889,6 +889,78 @@ Int_t Iact1dUnbinnedLkl::ReaddNdESignal(TString filename)
...
+  Double_t lowEdge  = 0.0;
+  Double_t highEdge = 0.0;
+  // open file and look for histograms
+  for(Int_t iFile=0;iFile<nFiles;iFile++)
+    {
+      dNdESignalFile[iFile]  = new TFile(filenames[iFile]);
+      hProvdNdESignalTemp = (TH1F*) dNdESignalFile[iFile]->Get("hdNdE");
+      // check that dN/dE histograms are found
+      if(!hProvdNdESignalTemp) status=1;
+      // store the histogram binning from the first historgram
```
javierrico: Fix typo

src/Iact1dUnbinnedLkl.cc
```diff
@@ -889,6 +889,78 @@ Int_t Iact1dUnbinnedLkl::ReaddNdESignal(TString filename)
...
+          nBins = hProvdNdESignalTemp->GetNbinsX();
+          lowEdge = hProvdNdESignalTemp->GetBinLowEdge(1);
+          highEdge = hProvdNdESignalTemp->GetBinLowEdge(nBins+1);
+        }
+      // check that the following histograms have the same binning than the first histogram
+      if(hProvdNdESignalTemp->GetNbinsX()!=nBins || hProvdNdESignalTemp->GetBinLowEdge(1)!=lowEdge || hProvdNdESignalTemp->GetBinLowEdge(nBins+1)!=highEdge)
+        {
+          cout << "Iact1dUnbinnedLkl::ReaddNdESignal, histogram read from file " << filenames[iFile] << " has " << hProvdNdESignalTemp->GetNbinsX() << " bins (should be " << nBins << " bins), lowEdge = " << hProvdNdESignalTemp->GetBinLowEdge(1) << " (should be " << lowEdge << ") and highEdge = " << hProvdNdESignalTemp->GetBinLowEdge(nBins+1) << " (should be " << highEdge << ")" << endl;
+        }
+      // interpolate if historgram binning divers from the binning of the first histogram
```
javierrico: Fix typo, I know, it's a hard word... :P 
Also, you interpolate always, independently of the binning of first histogram, so the comment is not correct

src/Iact1dUnbinnedLkl.cc
```diff
@@ -889,6 +889,78 @@ Int_t Iact1dUnbinnedLkl::ReaddNdESignal(TString filename)
...
+      // store the histogram binning from the first historgram
+      if(nBins==0)
+        {
+          nBins = hProvdNdESignalTemp->GetNbinsX();
+          lowEdge = hProvdNdESignalTemp->GetBinLowEdge(1);
+          highEdge = hProvdNdESignalTemp->GetBinLowEdge(nBins+1);
+        }
+      // check that the following histograms have the same binning than the first histogram
+      if(hProvdNdESignalTemp->GetNbinsX()!=nBins || hProvdNdESignalTemp->GetBinLowEdge(1)!=lowEdge || hProvdNdESignalTemp->GetBinLowEdge(nBins+1)!=highEdge)
+        {
```
javierrico: General style comment: do not use curly brackets when not needed (i.e. when they contain just one line)

src/Iact1dUnbinnedLkl.cc
```diff
@@ -889,6 +889,78 @@ Int_t Iact1dUnbinnedLkl::ReaddNdESignal(TString filename)
...
+      if(!hProvdNdESignalTemp) status=1;
+      // store the histogram binning from the first historgram
+      if(nBins==0)
+        {
+          nBins = hProvdNdESignalTemp->GetNbinsX();
+          lowEdge = hProvdNdESignalTemp->GetBinLowEdge(1);
+          highEdge = hProvdNdESignalTemp->GetBinLowEdge(nBins+1);
+        }
+      // check that the following histograms have the same binning than the first histogram
+      if(hProvdNdESignalTemp->GetNbinsX()!=nBins || hProvdNdESignalTemp->GetBinLowEdge(1)!=lowEdge || hProvdNdESignalTemp->GetBinLowEdge(nBins+1)!=highEdge)
```
javierrico: Since you are converting all hProvdNdESignalTemp to  histograms with exactly the same binning using readAndInterpolate, I don't think you need this check

src/Iact1dUnbinnedLkl.cc
```diff
@@ -889,6 +889,78 @@ Int_t Iact1dUnbinnedLkl::ReaddNdESignal(TString filename)
...
+  Double_t highEdge = 0.0;
+  // open file and look for histograms
+  for(Int_t iFile=0;iFile<nFiles;iFile++)
+    {
+      dNdESignalFile[iFile]  = new TFile(filenames[iFile]);
+      hProvdNdESignalTemp = (TH1F*) dNdESignalFile[iFile]->Get("hdNdE");
+      // check that dN/dE histograms are found
+      if(!hProvdNdESignalTemp) status=1;
+      // store the histogram binning from the first historgram
+      if(nBins==0)
```
javierrico: You don't need this, the output histos from readAndInterpolate should all have fNFineBins bins, from fFineLEMin to fFineLEMax (see other calls to the function). These values can be define by the user, but have sensible default values too, so you don't need to worry in your function to recompute them

src/Iact1dUnbinnedLkl.cc
```diff
@@ -889,6 +889,78 @@ Int_t Iact1dUnbinnedLkl::ReaddNdESignal(TString filename)
...
+      delete hProvdNdESignalTemp;
+    }
+
+  if(status==0)
+    {
+      // create histogram for the linear combination of dNdESignal
+      linCHdNdESignal = new TH1F("linCHdNdESignal","dN/dE for signal events",nBins,lowEdge,highEdge);
+      // perform the linear combination
+      for(Int_t iFile=0;iFile<nFiles;iFile++)
+        {
```
javierrico: I insist: do not use "{", "}" for single line loops

src/Iact1dUnbinnedLkl.cc
```diff
@@ -889,6 +889,78 @@ Int_t Iact1dUnbinnedLkl::ReaddNdESignal(TString filename)
...
+  Int_t nBins       = 0;
+  Double_t lowEdge  = 0.0;
+  Double_t highEdge = 0.0;
+  // open file and look for histograms
+  for(Int_t iFile=0;iFile<nFiles;iFile++)
+    {
+      dNdESignalFile[iFile]  = new TFile(filenames[iFile]);
+      hProvdNdESignalTemp = (TH1F*) dNdESignalFile[iFile]->Get("hdNdE");
+      // check that dN/dE histograms are found
+      if(!hProvdNdESignalTemp) status=1;
```
javierrico: Like this, if just the last file contains a valid histogram, the status is set to 1.
status should be 0 (and should be returned by the function) if ANY of the files contain no valid histogram

src/Iact1dUnbinnedLkl.cc
```diff
@@ -889,6 +889,78 @@ Int_t Iact1dUnbinnedLkl::ReaddNdESignal(TString filename)
...
+    }
+
+  if(status==0)
+    {
+      // create histogram for the linear combination of dNdESignal
+      linCHdNdESignal = new TH1F("linCHdNdESignal","dN/dE for signal events",nBins,lowEdge,highEdge);
+      // perform the linear combination
+      for(Int_t iFile=0;iFile<nFiles;iFile++)
+        {
+          linCHdNdESignal->Add(hProvdNdESignal[iFile],branchingRatios[iFile]);
```
javierrico: No check of whether BRs sum is 1 (or at least less than 1), should be included. Maybe you want to consider normalizing BRs to sum 1? A comment about what you do with this should be included in the comment at the top of the class definition.

javierrico (commented):
Another bunch!

src/Iact1dUnbinnedLkl.cc
tjarkmiener: Thank you for the additional comments! Yes, dNdESignalFile don't need to be an array of pointer. But hProvdNdESignal have to be an array of pointer. It's used in the second nFiles for loop at line 949.

src/Iact1dUnbinnedLkl.cc
tjarkmiener: Ok!

src/Iact1dUnbinnedLkl.cc
TjarkMiener: Yes, it's true. However, I thought it might be a good idea to have check before applying the interpolation as a warning for the users. Might be easier for the users to spot errors in their dN/dE histograms (if they don't take it from Cirelli). Should I remove this check?

src/Iact1dUnbinnedLkl.cc
tjarkmiener: Ok!

src/Iact1dUnbinnedLkl.cc
tjarkmiener: Yes!

src/Iact1dUnbinnedLkl.cc
tjarkmiener: Indeed it is ;P

src/Iact1dUnbinnedLkl.cc
tjarkmiener: I'm probably getting this wrong, but shouldn't be all histograms valid in order to be considered by the linear combination? Set status to 1 means that the file (or histogram) is not found, which I'm checking in line 921. I will add a check here. If branching ratio is 0.0, I will ignore the missing histogram and don't force gLike to break. If the histogram is not found and branching ratio != 0.0, I will set status to 1, which force gLike to break, which is necessary in my point of view. Please instruct if I'm wrong!

src/Iact1dUnbinnedLkl.cc
tjarkmiener: Good point! I will add a check!
Does the normalizing of BRs is linked to the comment above? Do you mean, if we can't find a histogram (ie with BR 0.3), we should normalize the rest 0.7 to 1?

src/Iact1dUnbinnedLkl.cc
javierrico: The second loop could be part of the first loop

src/Iact1dUnbinnedLkl.cc
javierrico: Ups, sorry, I always tend to think 0 is unsuccess, but it's 1, you are right...

src/Iact1dUnbinnedLkl.cc
javierrico: No,  if **any** of the requested histos is not found, then you should return the _unsuccess_ code (1, yes? right, 1 ;))

javierrico:
I posted some replies to your comments

src/Iact1dUnbinnedLkl.cc
tjarkmiener: Good point!

src/Iact1dUnbinnedLkl.cc
tjarkmiener: Yes :D

TjarkMiener:
I added the ability to select custom combinations of annihilation channels from the rc files. I also added a extra channel ('branon') and set a placeholder in the `./scripts/jointLklDM.C` script. I think it's the right place to calculate the BRs for the branon model, since they are mass dependent. Otherwise, I would have a rc file for each mass, which could be quite messy. What do you think of this structure?

rcfiles/jointLklDM.rc
```diff
@@ -35,14 +35,24 @@ jointLklDM.plotsDir: plots
...
+# channel to be analyzed (bb, tautau, mumu, WW, gammagamma, pi0pi0, pi0gamma, branon or custom)
+jointLklDM.Channel: custom
+
+# ONLY used if the channel 'custom' is selected. Otherwise, the lists will be ignored!
+# ChannelList contains a set of channels. 
+# Combinations of bb, tautau, mumu and WW are valid!
+# BRList contains the corresponding branching ratios.
+# The length of ChannelList and BRList have to be equal and the sum of the  BRList have to be1. 
+jointLklDM.ChannelList: bb tautau mumu WW
+jointLklDM.BRList: 1 0 0 0
```
javierrico: I think you should not add these two new keywords "ChannelList" and "BRList".
Instead, the allowed formats for keyword "Channel" should be expanded to accept things like:
  bb  
  0.5\*bb+0.2\*tautau+0.2\*gammagamma+0.1\*ee
  branon
Then you should modify jointLklDM to decode these string, check that unitarity of BR is not violated, and call SetdNdESignalFunction. 
By the way the old version of SetdNdESignalFunction can be removed and keep only your new (more general) version, which is also valid for 100% BR into a single channel
Pay special attention because the dNdE for some of the channels are passed as files containing histograms (e.g. bbar) with Iact1dUnbinned::ReaddNdESignal whereas others are passed as functions (like gammagamma or ee) with Iact1dUnbinned::SetdNdESignalFunction. 
Maybe you need a function that merges what is now done in  Iact1dUnbinned::ReaddNdESignal and Iact1dUnbinned::SetdNdESignalFunction, that can accept a number >=0 of functions and another number >= 0  of files containing histograms, each with its corresponding BR, and construct the overall histogram

scripts/jointLklDM.C
```diff
@@ -131,13 +131,32 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
   TString  fdNdEpSignalDir   = fInputDataPath+"/"+provval+"/";
   Float_t  mcG               = env->GetValue("jointLklDM.mcG",0.);  //assumed value of <sv> for simulations
   TString  massList          = env->GetValue("jointLklDM.MassList","");
+  TString  channelList       = env->GetValue("jointLklDM.ChannelList","");
+  TString  brList            = env->GetValue("jointLklDM.BRList","");
```
javierrico: Can be removed (see comment above)

scripts/jointLklDM.C
```diff
@@ -131,13 +131,32 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+  for(UInt_t iBr=0;iBr<nBrs;iBr++)
+      brval[iBr] = re[iBr].Atof();
+
+  // check that ChannelList and BRList have the same length
+  if(nChannels != nBrs)
+    {
+      cout << " ## Oops! The length of the ChannelList (" << nChannels << ") and BRList (" << nBrs << ") should be the same!  <---------------- FATAL ERROR!!!"<< endl;
+      return;
+    }
+
```
javierrico: All this needs to be changed so that the number of channels, channels, and their BRs are decoded from the "Channel" keyword

scripts/jointLklDM.C
```diff
@@ -158,6 +177,8 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
   else if(!channel.CompareTo("pi0pi0",TString::kIgnoreCase))     {strchannel = "#pi^{0}#pi^{0}";    minmass=0.135;}
   else if(!channel.CompareTo("gammapi0",TString::kIgnoreCase))   {strchannel = "#pi^{0}#gamma";     minmass=0.135/2.;}
   else if(!channel.CompareTo("pi0gamma",TString::kIgnoreCase))   {strchannel = "#pi^{0}#gamma";     minmass=0.135/2.;}
+  else if(!channel.CompareTo("branon",TString::kIgnoreCase))     {strchannel = "#branon";           minmass=0;}
+  else if(!channel.CompareTo("custom",TString::kIgnoreCase))     {strchannel = "#custom";           minmass=0;}
```
javierrico: "custom" can be removed, since with my proposal all allowed values of "Channel" keyword should be custom

scripts/jointLklDM.C
```diff
@@ -158,6 +177,8 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
   else if(!channel.CompareTo("pi0pi0",TString::kIgnoreCase))     {strchannel = "#pi^{0}#pi^{0}";    minmass=0.135;}
   else if(!channel.CompareTo("gammapi0",TString::kIgnoreCase))   {strchannel = "#pi^{0}#gamma";     minmass=0.135/2.;}
   else if(!channel.CompareTo("pi0gamma",TString::kIgnoreCase))   {strchannel = "#pi^{0}#gamma";     minmass=0.135/2.;}
+  else if(!channel.CompareTo("branon",TString::kIgnoreCase))     {strchannel = "#branon";           minmass=0;}
```
javierrico: the "#" symbol is to indicate a latex kind of symbol (e.g, "#pi" is like "\pi" in latex), so no need to use it here

scripts/jointLklDM.C
```diff
@@ -167,26 +188,39 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+  cout << "  * Channel values        : "; 
+  for(UInt_t iChannel=0;iChannel<nChannels;iChannel++)
+    cout << channelval[iChannel] << ((iChannel<nChannels-1)? ", " : "");
+  cout << endl;
+  // how many and which branching ratio values are we considering, if 'custom' channel is selected?
+  cout << " ** NUMBER OF BRS         : " << nBrs << endl;
+  cout << "  * Branching ratio values: "; 
+  for(UInt_t iBr=0;iBr<nBrs;iBr++)
+    cout << brval[iBr] << ((iBr<nBrs-1)? ", " : "");
+  cout << endl;
```
javierrico: Revise this block according to the new way of specifying the channel

scripts/jointLklDM.C
```diff
@@ -451,20 +485,64 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+
+                    // Try to read dNdE from existing file
+                    if(fullLkl->ReaddNdESignal(dNdESignalFileName))
+                      {
+                        cout << "Failed! <---------------- FATAL ERROR!!!" << endl;
+                        return;
+                      }
+                    else
+                      cout << "Ok!" << endl;
+                  }
```
javierrico, line 530: Also all this part needs to be redone to adapt to my first comment of this review

scripts/testBranonDMSearches.C
```diff
@@ -22,10 +22,10 @@ void testBranonDMSearches()
```
javierrico: testBranonDMSearches.C should not be committed to the main branch

javierrico (changes requested):

rcfiles/jointLklDM.rc
tjarkmiener: Thanks for your suggestion. Having these strings is more user-friendly and less error-prone. I already adopted the format and modified jointLklDM to decode these strings. 
I will try to tackle the issue with SetdNdESignalFunction in the next days!

tjarkmiener: I spent some time to get familiar with the `Iact1dUnbinned::SetdNdESignalFunction`/ `Iact1dUnbinned::AdddNdESignalFunction` and the structure of calling these function from the jointLklCM.C script. I was thinking about a strategy to add the multichannel feature, but is a bit tricky to find a good solution. A good and clean approach (in my point of view) to solve this problem is to move some parts of the jointLklDM.C into the Iact1dUnbinned class. Since this will change the structure and the arguments of the function, I want to suggest it to you before I start implementing it. Maybe I'm missing some important points, which you are aware of.

My suggestion:
- We are calling a new `Iact1dUnbinned::SetdNdESignalFunction` with the arguments (pointer to a list of channels, pointer to a list of the corresponding branching ratios, mass value "mdm").
This call would be also valid when the lists only contain one element (100% BR into single channel)
- We will move the channel discrimination in the Iact1dUnbinned class and call the `SetdNdESignalFunction` from there. Therefore the user, who is likely interacting on rc configuration file- and script-level, don't have to worry about setting up the right `SetdNdESignalFunction` with the right parameters for each channel. The user only have to call one function with a channel and branching ratio list plus the mass value "mdm" and it will automatically set up the right dN/dE histogram for them.

I think it's a clean and intuitive way to tackle this problem. Please let me know if you have any doubts.

javierrico: The problem with this approach is that Iact1dUnbinned should be agnostic about the physics behind the signal spectra it's looking for. 
What is the difference between your proposal and doing exactly the same thing in the script?
My idea is that all scripts should become executables at some point and user should not be allowed to change them. Users should only touch the rc files.
Is there any fundamental reason why the interpretation of the setting up of the dNdE histogram cannot happen in the script?
If there really is, I am open to include it in the class, but we need a new class which inherits from Iact1dUnbinned and contains the DM-related part. I could take care of it, because some parts of the current Iact1dUnbinned class should go to the new one. But let's see first if it's really necessary

tjarkmiener: Oh I see! I forgot the point that gLike is a general tool and not specific for DM searches! It makes totally sense to have every DM relative things in the DM script and not in the general Unbinned lkl class. My concern was only based on code readability and compactness.

tjarkmiener: This is solved with the latest commit.

TjarkMiener:
I tested the changes with 100% annihilation into each channel. I compared the results with the current master and they were identical.

rcfiles/jointLklDM.rc
```diff
@@ -35,14 +35,18 @@ jointLklDM.plotsDir: plots
 # Comment out if you don't want to save or read the histograms but create them on the fly
 # (that is slower but safer because you make sure you use the proper histograms and not
 #  some pre-existing version created for a different set of IRFs or samples)
-jointLklDM.dNdEpSignalDir: DM/dNdEpSignal
+# Comment out for the channel 'branon' for now!!!
+# Comment out for the channel 'custom' always!!!
```
javierrico: I don't think these two comments are necessary, for branon the total dN/dE is fixed right, so it makes perfect sense to allow writing the resolution-convoluted dNdEp histograms. 
Custom should not be a channel, but rather a given "br1*channel1+...+brn*channeln" string, which I think can be used in the name of the dNdEp file without any problem (maybe some "\" commands have to be included, no big deal...)

rcfiles/jointLklDM.rc
```diff
@@ -35,14 +35,18 @@ jointLklDM.plotsDir: plots
...
 ######################
 ANALYSIS CONFIGURATION
 ######################
 
-# channel to be analyzed (bb, tautau, mumu, WW, gammagamma, pi0pi0 or pi0gamma)
-jointLklDM.Channel: bb
+# channel to be analyzed (bb, tautau, mumu, WW, gammagamma, pi0pi0, pi0gamma, ee, branon or an arbitrary linear combination of them)
+# The linear combination must be written this form:'br_1*channel_1+...+br_n*channel_n'!
+jointLklDM.Channel: 1.0*bb+0.0*tautau
+#jointLklDM.Channel: bb
```
javierrico: Remove this line, single BRs an channels should be a particular case of the general, multiple BRs and channels. 
You have to allow that no BR value is provided for a given channel, in which case it should be interpreted as BR=1.

scripts/jointLklDM.C
```diff
@@ -158,7 +160,9 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
   else if(!channel.CompareTo("pi0pi0",TString::kIgnoreCase))     {strchannel = "#pi^{0}#pi^{0}";    minmass=0.135;}
   else if(!channel.CompareTo("gammapi0",TString::kIgnoreCase))   {strchannel = "#pi^{0}#gamma";     minmass=0.135/2.;}
   else if(!channel.CompareTo("pi0gamma",TString::kIgnoreCase))   {strchannel = "#pi^{0}#gamma";     minmass=0.135/2.;}
-  else strchannel = "";
+  else if(!channel.CompareTo("branon",TString::kIgnoreCase))     {strchannel = "branon";            minmass=0;}
+  else                                                           {strchannel = channel;             minmass=0;}
```
javierrico: channel is the string you read from the rcfile, and the one you should decode to build the dNdE histogram and to write the strchannel, that should be the latex version of channel (with greeks letters, etc) that it is used for plots. 
Please do not use strchannel for the decoding part as you do below
Now that channel can be any arbitrary string, you should write a separate function that would take channel as argument, and gives back the corresponding strchannel 
Include in the new function the option for channel="ee", that is now missing. It should be strchannel="e^{+}e^{-}"

scripts/jointLklDM.C
```diff
@@ -158,7 +160,9 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+  TString channellist = "bb_tautau_mumu_WW_gammagamma_pi0pi0_gammapi0_pi0gamma_ee_branon";
```
javierrico: This is not necessary, there should be no special treatment for the single-BR, single-channel case, it should be just a particular case of the general multiple-BR, multiple-channel case

scripts/jointLklDM.C
```diff
@@ -167,26 +171,85 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+  cout << "*** CHANNEL                  : " << channel << endl;
+  Int_t nChannels = 1;
+  TString* channelval = new TString[nChannels];
+  Double_t* brval = new Double_t[nChannels];
+  if(channellist.Contains(channel))
+    {
+       channelval[0] = channel;
+       brval[0] = 1.0;
+    }
+  else
```
javierrico: Everything enclosed in this "else" should be a separate function, and also valid for nChannels=1, so that what is after the "if" clause can be removed. The function should give back the number of channels, the list of BRs (normalized) values and the list of pure channels they refer to.

scripts/jointLklDM.C
```diff
@@ -167,26 +171,85 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+      TObjArray *coefficients = strchannel.Tokenize("+");
+      nChannels = coefficients->GetEntries();
+      channelval = new TString[nChannels];
+      brval = new Double_t[nChannels];
+      TString factorstring, coefficientstring;
+      for (Int_t iChannel = 0; iChannel < nChannels; iChannel++) 
+        {
+          factorstring = (TString)((TObjString *)(coefficients->At(iChannel)))->String();
+          TObjArray *factors = factorstring.Tokenize("*");
+          if (factors->GetEntries() != 2)
```
javierrico: You should allow for number of entries to be 1, in case BR=1, this will make (e.g.) the string "bb" be valid, so that you do not need a separate code for the nChannels=1 case

scripts/jointLklDM.C
```diff
@@ -167,26 +171,85 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+  if(channellist.Contains(channel))
+    {
+       channelval[0] = channel;
+       brval[0] = 1.0;
+    }
+  else
+    {
+      TObjArray *coefficients = strchannel.Tokenize("+");
+      nChannels = coefficients->GetEntries();
+      channelval = new TString[nChannels];
```
javierrico: MEMORY LEAK! do not redirect the pointer without releasing the memory of the previous one!

scripts/jointLklDM.C
```diff
@@ -167,26 +171,85 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+      brval = new Double_t[nChannels];
```
javierrico: MEMORY LEAK! do not redirect the pointer without releasing the memory of the previous one!

scripts/jointLklDM.C
```diff
@@ -167,26 +171,85 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+              if (coefficientstring.IsFloat() && i == 0)
+                brval[iChannel] = coefficientstring.Atof();
+              else if(channellist.Contains(coefficientstring) && i == 1)
+                channelval[iChannel] = coefficientstring;
+              else
+                {
+                  cout << " ## Oops! Something went wrong with the parsing of the channel (" << channel << ")!  <---------------- FATAL ERROR!!!"<< endl;
+                  return;
+                }
+              TString coefficientstring = (TString)((TObjString *)(factors->At(i)))->String();
```
javierrico: This object was already declared, line probably can be safely removed

scripts/jointLklDM.C
```diff
@@ -167,26 +171,85 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+            }
+        }
+      // how many and which channel values are we considering, if a custom linear combination of channels is selected?
+      cout << " ** Custom linear combination:" << endl;
+      cout << "  * Number of coefficients   : " << nChannels << endl;
+      cout << "  * Channel values           : ";
```
javierrico: These three lines should be replaced by one, just reporting the channel as a string of the type "br1*channel1+...+brn*channeln"
It is important to not just return channel, but to write it again with the values of nChannels, brval and channelval, previously decoded, in this way you can spot mistakes in your decoding function

javierrico (changes requested):
This is progressing in the good direction, but please consider this new bunch of comments.
Also, as a general comment, try to provide more comments in the code. 
Cheers

rcfiles/jointLklDM.rc
tjarkmiener: This comments are remains from the time I used the 'custom' as a channel keyword. I will delete them.

scripts/jointLklDM.C
tjarkmiener: True!

scripts/jointLklDM.C
tjarkmiener: True!

scripts/jointLklDM.C
tjarkmiener: Ok!

scripts/jointLklDM.C
tjarkmiener: Ok!

scripts/jointLklDM.C
tjarkmiener: New commit does!

scripts/jointLklDM.C
tjarkmiener: New commit does!

scripts/jointLklDM.C
tjarkmiener: Ok!

scripts/jointLklDM.C
tjarkmiener: New commit does! Added the "ee" channel.

rcfiles/jointLklDM.rc
tjarkmiener: Yes!

TjarkMiener:
Hi Javier, I was able to produce the first preliminary branon limits with the current available channels in gLike. Next week, I will tackle #8 and add the missing channels (we also need some of them for the gloryduck project). One cool thing, which is also on my ToDos is the reproduction of the [Figure 1](https://arxiv.org/pdf/1709.09819.pdf) as a crosscheck. So far it's seems to work as it should be and I'm getting reasonable branching ratios! Cheers

TjarkMiener:
I created the missing dN/dE histograms using the two scripts you sent me. Is there a cross-check needed?

When I'm selecting masses below 70 GeV the code is breaking and we aren’t sure why. There is a warning saying "fHdNdESignal is not normalized!". Probably we don't want to go to masses below 70 GeV, but we are still worried about the break. Do you have an idea what caused the break? 

 *** End of reading dN_signal/dE and dN_signal/dE' histograms
Iact1dBinnedLkl::MakeChecks (Iact1dBinnedLkl_02) Warning: fHdNdESignal is not normalized!
 *** Computing -2logL (parabola) vs g:
Lkl::ComputeLklVsG (JointLkl_00) Message: Finding minimum of -2logL... 
JointLkl::ReorderSamples (JointLkl_00) Message: sample Iact1dUnbinnedLkl_01 has been sent to the end
JointLkl::ReorderSamples (JointLkl_00) Message: sample Iact1dBinnedLkl_02 has been sent to the end
JointLkl::ReorderSamples (JointLkl_00) Warning: none of the samples will produce any signal event
Lkl::MinimizeLkl (JointLkl_00) Warning: checks were not successfull!!!!

 *** Break *** segmentation violation

scripts/jointLklDM.C
```diff
@@ -147,18 +151,40 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+      else if(!channelval[iChannel].CompareTo("WW",TString::kIgnoreCase))         {strchannel.Append("W^{+}W^{-}");        minmass=80.3;}
+      else if(!channelval[iChannel].CompareTo("ZZ",TString::kIgnoreCase))         {strchannel.Append("Z^{+}Z^{-}");        minmass=91.2;}
+      else if(!channelval[iChannel].CompareTo("hh",TString::kIgnoreCase))         {strchannel.Append("hh");                minmass=125;}
+      else if(!channelval[iChannel].CompareTo("gammagamma",TString::kIgnoreCase)) {strchannel.Append("#gamma#gamma");      minmass=0;}
+      else if(!channelval[iChannel].CompareTo("pi0pi0",TString::kIgnoreCase))     {strchannel.Append("#pi^{0}#pi^{0}");    minmass=0.135;}
+      else if(!channelval[iChannel].CompareTo("gammapi0",TString::kIgnoreCase))   {strchannel.Append("#pi^{0}#gamma");     minmass=0.135/2.;}
+      else if(!channelval[iChannel].CompareTo("pi0gamma",TString::kIgnoreCase))   {strchannel.Append("#pi^{0}#gamma");     minmass=0.135/2.;}
+      else if(!channelval[iChannel].CompareTo("ee",TString::kIgnoreCase))         {strchannel.Append("e^{+}e^{-}");        minmass=0.511;}
```
javierrico: electron mass is 511*e-6 [GeV]

scripts/jointLklDM.C
```diff
@@ -810,3 +866,98 @@ void setDefaultStyle()
...
+              return;
+            }
+          // set the branching ratio to 1.0 (100%), since there is only one channel selected
+          brval[iChannel] = sumBR = 1.0;
+        }
+      // if there is a linear combination of channels selected
+      else
+        {
+          // check if the channel argument is in the right form
+          if (factors->GetEntries() != 2)
```
javierrico: Since you normalize at the end, I think strings like "bb+0.5*tt" should be allowed

scripts/jointLklDM.C
```diff
@@ -147,18 +151,40 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+      else if(!channelval[iChannel].CompareTo("ZZ",TString::kIgnoreCase))         {strchannel.Append("Z^{+}Z^{-}");        minmass=91.2;}
+      else if(!channelval[iChannel].CompareTo("hh",TString::kIgnoreCase))         {strchannel.Append("hh");                minmass=125;}
+      else if(!channelval[iChannel].CompareTo("gammagamma",TString::kIgnoreCase)) {strchannel.Append("#gamma#gamma");      minmass=0;}
+      else if(!channelval[iChannel].CompareTo("pi0pi0",TString::kIgnoreCase))     {strchannel.Append("#pi^{0}#pi^{0}");    minmass=0.135;}
+      else if(!channelval[iChannel].CompareTo("gammapi0",TString::kIgnoreCase))   {strchannel.Append("#pi^{0}#gamma");     minmass=0.135/2.;}
+      else if(!channelval[iChannel].CompareTo("pi0gamma",TString::kIgnoreCase))   {strchannel.Append("#pi^{0}#gamma");     minmass=0.135/2.;}
+      else if(!channelval[iChannel].CompareTo("ee",TString::kIgnoreCase))         {strchannel.Append("e^{+}e^{-}");        minmass=0.511;}
+      else if(!channelval[iChannel].CompareTo("branon",TString::kIgnoreCase))     {strchannel.Append("branon");            minmass=0;}
+      else strchannel = "";
+    }
```
javierrico, line 199: I think the "*" sign is missing, is that on purpose?

scripts/jointLklDM.C
```diff
@@ -147,18 +151,40 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+  Int_t nChannels = coefficients->GetEntries();
+  TString* channelval = new TString[nChannels];
+  Double_t* brval = new Double_t[nChannels];
+  // call the function that decodes the coefficients of the channel
+  decode_channel(coefficients,nChannels,channelval,brval);
+
+  // annihilation/decay channel string
   TString  strchannel;
   Double_t minmass = 0;
-  if     (!channel.CompareTo("bb",TString::kIgnoreCase))         {strchannel = "b #bar{b}";         minmass=5;}
```
javierrico, line 153: Later, on line 528, channel is used directly to build the dNdEpSignalFileName, but this is not ok because one should use the equivalent with normalized BRs

javierrico (changes requested):
Apart from this comments, please make a check that for the new created dNdE histos you get something reasonable (e.g. comparing it with results in the literature, or with Cembranos parameterization)

javierrico:
Hi Tjark
Your latest commits seem 99.9% ok, I have just added some comments that I hope can be implemented relatively fast, and then probably we are ready for the merge.
I am not able to reproduce your problem for mass 70 GeV, can you send me the exact rcfile and code version?

scripts/jointLklDM.C
tjarkmiener: Yes!

scripts/jointLklDM.C
tjarkmiener: OK

scripts/jointLklDM.C
tjarkmiener: Probably adding a "\cdot" would be suited well?

scripts/jointLklDM.C
tjarkmiener: Good catch!

scripts/jointLklDM.C
tjarkmiener: Or should it be "#cdot"? I'm confused by the "# \" convention.

javierrico: Try #cdot, but check the result since I'm not sure

tjarkmiener: "#upoint" or "#bullet" is doing the job :)

TjarkMiener:
It's with the current code in this branon branch, when selecting masses like 10 or 50 Gev (in the common "jointLklDM.rc").

scripts/jointLklDM.C
```diff
@@ -579,8 +578,10 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
 		hadcanvas[isample]->SetName(Form("hadcanvas_%d",isample));
 		hadcanvas[isample]->SetTitle(Form("IRFs and data for sample %d",isample));
 		hadcanvas[isample]->cd(5);
-		TLatex* ltchannel = new TLatex(0.8,0.8,strchannel);
-		ltchannel->SetTextSize(0.055);
+        TLatex* ltchannel;
+        if(nChannels == 1) ltchannel = new TLatex(0.8,0.8,strchannel);
+        else ltchannel = new TLatex(0.7,0.8,strchannel);
+        ltchannel->SetTextSize(0.055);
```
javierrico: Please indent

javierrico (changes requested):
Ok, just a tiny minor thing (indentation...) and we're done, sorry for being so strict...
Once you commit that I will merge, without waiting for anything related to the comments below

About getting a crash for low masses as 10 and 50 GeV that is "normal" because there is no sensitivity with MAGIC to those masses. More precisely, when you take the dN/dE for those masses, multiply by effective area, and convolute with the energy dispersion and integrate over the minimum reconstructed energy value considered reliable (~60 GeV if I remember correctly) you get 0.
It is not nice that the program crashes, and I will put in my todo list to treat this case more cleanly, but for the time being just do not try to use MAGIC data for masses that do not produce gammas in the MAGIC sensitive range.

Final comment, please check the new dNdE histos you have added by plotting a few of them and comparing with the literature. I checked fast and they look ok, but probably a more careful check is needed.

javierrico:
Also, I did the check of using channel = bb+0.5\*tautau and I get:
*** CHANNEL                  : bb+0.5\*tautau
 ** Channel check            : 0.6\*bb+0.3\*tautau
there could be a problem with normalization or with the rounding of the BRs when writing them, which would be confusing, can you please check also that?

scripts/jointLklDM.C
tjarkmiener: The editor messed that up. Should be fixed with the next commit.

TjarkMiener:
To compute the dN/dE histos is the exact br values are used. That problem only occurred, when displaying the channel (on console and in the plots). Back then, I didn't found a handy rounding function in TMath, so I was flooring/cutting the exact br values. The next commit is solving this problem. However, the dN/dE histos is still calculated with the exact br values, which should be the right way to go.

Thanks for your explanation about the crash. This makes sense to me and this is what we guessed.

I will have a closer check for the dN/dE histos and come back to you here.

PR #12, opened 2019-05-03
Adding the possibility to export data to the Glory Duck format.

dkerszberg:


rcfiles/jointLklDM.rc
```diff
@@ -21,6 +21,13 @@ jointLklDM.plotmax: 1e-22
 # scale factor for results (e.g. to simulate for a Tobs and get results for another Tobs)
 # jointLklDM.plotScale: 1
 
+# Configuration for data exporter
```
javierrico: Please add a bit more of information in the comment, data exporter from what to what? and also on the other variables below

rcfiles/jointLklDM.rc
```diff
@@ -21,6 +21,13 @@ jointLklDM.plotmax: 1e-22
...
+jointLklDM.exportData: FALSE
+jointLklDM.svMin: 1e-28
+jointLklDM.svMax: 1e-18
+jointLklDM.svNPoints: 1000
+jointLklDM.svLogStep: TRUE
+jointLklDM.logJ: 19.
```
javierrico: It still bothers me to have the values for a given logJ value. Is it possible still to consider tabulating lkl vs <sv>*J, that would be cleaner, because there is no need of matching J factors in this line and below.

Another comment, around line 92 of the rc file you should document the syntax of the input string specific for the GloryDuckTables2019Lkl samples, as it is done for the rest of the possible likelihood classes

rcfiles/jointLklDM.rc
```diff
@@ -21,6 +21,13 @@ jointLklDM.plotmax: 1e-22
 # scale factor for results (e.g. to simulate for a Tobs and get results for another Tobs)
 # jointLklDM.plotScale: 1
 
+# Configuration for data exporter
+jointLklDM.exportData: FALSE
+jointLklDM.svMin: 1e-28
```
javierrico: Do not call these variables "sv"-something, but include "export", or "GD", to differentiate them from other possible uses

scripts/jointLklDM.C
```diff
@@ -131,7 +134,13 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
   TString  fdNdEpSignalDir   = fInputDataPath+"/"+provval+"/";
   Float_t  mcG               = env->GetValue("jointLklDM.mcG",0.);  //assumed value of <sv> for simulations
   TString  massList          = env->GetValue("jointLklDM.MassList","");
-        
+  Bool_t   exportData        = env->GetValue("jointLklDM.exportData",kFALSE);
+  Double_t svMin             = env->GetValue("jointLklDM.svMin",0.);
+  Double_t svMax             = env->GetValue("jointLklDM.svMax",0.);
+  Bool_t   svLogStep         = env->GetValue("jointLklDM.svLogStep",kFALSE);
```
javierrico: Probably default should actually be log step, right?

scripts/jointLklDM.C
```diff
@@ -165,7 +174,44 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+
+  // Exporting data to Glory Duck format
+  Double_t svStep = 0.;
+  Double_t isv = svMax;
+  Double_t svScan[svNPoints+1];
+  vector<vector<Double_t> > vlkl2D(svNPoints+1);
+  if (exportData)
+    {
+      // Create directory and open file for data export
+      gSystem->Exec(Form("mkdir -p %s/Data/txt",fPlotsDir.Data()));
```
javierrico: - weird directory to store results, I suggest to have specific variable in rc file to specify this
- why hardcode "Data"? this can also be done for simulations (will be actually needed for the brazilian bands)

scripts/jointLklDM.C
```diff
@@ -165,7 +174,44 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
   Int_t           nskippedmass = GetNSkippedMasses(nmass0,massval0,minmass);
   Int_t           nmass        = nmass0-nskippedmass; 
   const Double_t* massval      = massval0+nskippedmass;
-  
+
+  // Create stream to export data
+  std::ofstream data;
+
+  // Exporting data to Glory Duck format
```
javierrico: Mind the comment: This is not where the actual exportation happens, right? just preparing some stuff for that exportation.

scripts/jointLklDM.C
```diff
@@ -187,6 +233,7 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
   cout << "*** IRF/DATA PLOTS      : " << (showSamplePlots?   "YES" : "NO") << endl;
   cout << "*** PARABOlA PLOTS      : " << (showParabolaPlots? "YES" : "NO") << endl;
   cout << "*** Plot y-axis range   : " << plotmin << " to " << plotmax << Form(" %s", (isDecay? "s" : "cm^3/s")) << endl;
+  cout << "*** EXPORT DATA         : " << (exportData? "YES" : "NO") << endl;
```
javierrico, line 244: Also mention here export to what

scripts/jointLklDM.C
```diff
@@ -301,6 +348,30 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+                  if(TMath::Abs(massToBeTested-massInTheFile) < 1.e-6)
+                    massAvailable = kTRUE;
+                }
+            }
+          if(!massAvailable)
+            {
+              cout << " ## Oops! At least one mass to be tested is not in the file given as an input <---------------- FATAL ERROR!!!"<< endl;
+              return;
+            }
+          delete tmpLkl;
```
javierrico: why deleting and recreating?
why not just 
lkl[iLkl] = tmpLkl
?

scripts/jointLklDM.C
```diff
@@ -373,142 +445,162 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+		  delete hdNdEpSignal;
+		  cout << "Ok!" << endl;
+	        } 
+	    }
+          else if(!strcmp(lkl[isample]->ClassName(),"GloryDuckTables2019Lkl"))
+            {
+              GloryDuckTables2019Lkl* gdLkl = dynamic_cast<GloryDuckTables2019Lkl*>(lkl[isample]);
+              if(gdLkl->SetActiveMass(mass))
+                {
+                  cout << "Failed! <---------------- FATAL ERROR!!!" << endl;
```
javierrico: Add a more informative error message (follow similar examples for other errors)

scripts/jointLklDM.C
```diff
@@ -373,142 +445,162 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+                  return;
+                } 
+            }
+        } // end of loop over samples
       cout << " *** End of reading dN_signal/dE and dN_signal/dE' histograms" << endl;  
 
+      // Case where dataset don't contain signal event
```
javierrico: Why this check here? 
wasn't it there before?

scripts/jointLklDM.C
```diff
@@ -517,7 +609,7 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
 	  {
 	    Iact1dUnbinnedLkl* fullLkl = dynamic_cast<Iact1dUnbinnedLkl*>(sample[isample]);
 	    
-	    if(imass==0)
+	    if(!Init_canvas_samples)
```
javierrico, line 731: What was the problem motivating this change?

scripts/jointLklDM.C
```diff
@@ -624,10 +718,34 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
 	  gPad->Modified();
 	  gPad->Update();
 	}
-      
+
+      // Save -2logLkl vs <sv> in file
+      //////////////////////////////////////////////////////////////
+      if (exportData)
+        {
```
javierrico: To keep homogeneous style: do not use bracket here, only when strictly necessary

scripts/jointLklDM.C
```diff
@@ -624,10 +718,34 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+          for (Int_t isv=0; isv<=svNPoints; isv++)
+            vlkl2D[isv][imass] = grLklParabola[imass]->Eval(svScan[isv]) - grLklParabola[imass]->Eval(0);
+        }
+
     } // end of loop over DM masses
-  
-  
+
+    if (exportData)
+      {
```
javierrico, line 862: To keep homogeneous style: do not use bracket here, only when strictly necessary

src/GloryDuckTables2019Lkl.cc
```diff
@@ -232,7 +236,11 @@ Int_t GloryDuckTables2019Lkl::ReadGloryDuckInputData(TString filename)
     {
       if(read_logJ)
         {
-          logJ = field;
+          if(TMath::Abs(field-fLogJ) > 1.e-6)
```
javierrico, line 241: As I said, I think the problem can be solved if the values of lkl are stored as a function of \<sv\>*J
In any case, if we know what J is used to produce the values of lkl vs <sv>, you can compute them for any other J, since \<sv\> and J are degenerate in the flux computation. I would therefore not issue and error but a warning and then compute properly the values of lkl taking into account the difference between the J in the rcfile and the one in the data file

scripts/jointLklDM.C
```diff
@@ -301,6 +348,30 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
 	}
+      else if(classType.CompareTo("GloryDuckTables2019Lkl")==0)
+	{
+	  GloryDuckTables2019Lkl *tmpLkl =  new GloryDuckTables2019Lkl(inputString);
+          Bool_t massAvailable = kFALSE;
+          for(Int_t jmass=0;jmass<nmass;jmass++)
+            {
+              massAvailable = kFALSE;
+              Double_t massToBeTested = massval[jmass];
+              for(Int_t kmass=0;kmass<tmpLkl->GetNMasses();kmass++)
```
javierrico: This produces a warning, please fix:
```
/Users/jrico/Dropbox/DM/gLike/./scripts/jointLklDM.C:359:38: warning: comparison of integers of different signs: 'Int_t' (aka 'int') and 'UInt_t' (aka 'unsigned int') [-Wsign-compare]
              for(Int_t kmass=0;kmass<tmpLkl->GetNMasses();kmass++)
                                ~~~~~^~~~~~~~~~~~~~~~~~~~~

```

scripts/jointLklDM.C
```diff
@@ -517,7 +609,7 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
 	  {
 	    Iact1dUnbinnedLkl* fullLkl = dynamic_cast<Iact1dUnbinnedLkl*>(sample[isample]);
 	    
-	    if(imass==0)
+	    if(!Init_canvas_samples)
 	      {
 		hadcanvas[isample] = fullLkl->PlotHistosAndData();
```
javierrico, line 733: Something weird happens now. I have run the default rc file, activating the creation of the GD files, for masses 100, 1000, 10000 and 100000
When I look at the sample plots, the dN/dE for sample 1 and masses 10000 and 100000 are lower than they should. 
It doesn't seem to have any effect on the limits, but should be fixed.
No idea where the problem is, but probably you can find faster. 
I include the sample 1 plots for the master branch and your glory branch, so that you see the difference
[test_Data_histos_sample01.pdf](https://github.com/javierrico/gLike/files/3450720/test_Data_histos_sample01.pdf)
[testGloryDuck_Data_histos_sample01.pdf](https://github.com/javierrico/gLike/files/3450721/testGloryDuck_Data_histos_sample01.pdf)

javierrico (changes requested):

rcfiles/jointLklDM.rc
dkerszberg: Done.

rcfiles/jointLklDM.rc
dkerszberg: 1st part of the comment: I rememeber the conversation and I agree. Before implementing anything I would prefer we submit the idea ti the rest of the GD group to get their feedback. Also keeping this version (hopefully just for now) would allow me to continue to work with the ICRC files. No problem on my side to keep this PR open until this question is solved.

2nd part of the comment: done.

rcfiles/jointLklDM.rc
dkerszberg: Done.

scripts/jointLklDM.C
dkerszberg: Indeed, fixed.

scripts/jointLklDM.C
dkerszberg: Fixed.

scripts/jointLklDM.C
dkerszberg: Done.

scripts/jointLklDM.C
dkerszberg: Done.

src/GloryDuckTables2019Lkl.cc
dkerszberg: Agreed (see my answer above) but I didn't touch it for now. Just to be sure: in that case you would compute the lkl values that are correct for the J-factor of the rcfile, ie the one we want to use for combination, using the J-factor written in the file as "start point" (because it is the one used to compute the lkl actually stored in the file).

scripts/jointLklDM.C
dkerszberg: Fixed. I also took the opportunity to remove another warning about fullLkl that would be used without being initialized (so I initialize it to NULL) that I had but I don't know if you also had it...

scripts/jointLklDM.C
dkerszberg: Fixed. This was linked to the way I was taking care of mass for which some instruments don't have any data. I had to rewrite that part.

scripts/jointLklDM.C
dkerszberg: No idea... Fixed!

scripts/jointLklDM.C
dkerszberg: Done.

scripts/jointLklDM.C
dkerszberg: There was no such check no. This is why if we were running an analysis for a mass for which no data are available the program just crashed. But this method was causing some issue with the plots as you noted in another comment so I change the approach here for something else.

rcfiles/jointLklDM.rc
javierrico: Let me suggest:

```
# Configuration for data exporter into the Glory Duck file format (i.e. likelihood values for all DM masses specified below)
jointLklDM.exportData: TRUE
# minimum, maximum and number of scanned <sv> values
jointLklDM.exportSvMin: 1e-28
jointLklDM.exportSvMax: 1e-18
jointLklDM.exportSvNPoints: 1000
# log-spaced <sv> values? 
jointLklDM.exportSvLogStep: TRUE
# Assumed log10(Jfactor) when computing <sv>
jointLklDM.exportLogJ: 19.
# Directory for output files
jointLklDM.exportDataPath: exported_files/
```

javierrico: Also, even if we save the logJ into the file, why do you need to specify it here?
This conversion only makes sense if all lkl terms at the end of the rc file refer to the same J-factor, right? so maybe one could check that all J-factors from the different terms are the same before producing the output with such J-factor written in the output file. Also one could add the check that no DeltaJ different from zero is included

scripts/jointLklDM.C
javierrico: Can you also fix the indentation of this if(exportData) block and the next one in line 731? The "for" lines are too many spaces to the right, and so it is the second of these "if"s

src/GloryDuckTables2019Lkl.cc
javierrico: Yes, that is what I mean. In this way you are allowed to compute <sv> values for any J-factor, and not just the one written in the file. 
I think this you can do without the agreement of anyone from the GD group, this is just an internal working rule of gLike

scripts/jointLklDM.C
javierrico: Can you comment on this please?

scripts/jointLklDM.C
```diff
@@ -648,7 +643,13 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
       
       // compute -2logLkl vs g for precise limit computation
       cout << " *** Computing -2logL (parabola) vs g:" << endl;
-      lkl[0]->ComputeLklVsG();
+      if(!lkl[0]->ComputeLklVsG())
+        {
+          cout << " *** Skipping DM mass = " << mass << " GeV because checks were not successfull (maybe none of the samples will produce any signal event?)" << endl;
+          svLimVal[imass] = 0.;
```
javierrico, line 778: These values should be set to something very big, not zero. Like 9e99 or so...

scripts/jointLklDM.C
```diff
@@ -722,15 +723,12 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
       // Save -2logLkl vs <sv> in file
       //////////////////////////////////////////////////////////////
       if (exportData)
-        {
           for (Int_t isv=0; isv<=svNPoints; isv++)
```
javierrico: Move this 2 [?] spaces to the left

scripts/jointLklDM.C
```diff
@@ -722,15 +723,12 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
             vlkl2D[isv][imass] = grLklParabola[imass]->Eval(svScan[isv]) - grLklParabola[imass]->Eval(0);
-        }
 
     } // end of loop over DM masses
 
     if (exportData)
```
javierrico: Move this 2 [?] spaces to the left

scripts/jointLklDM.C
```diff
@@ -722,15 +723,12 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
-      {
         for (Int_t isv=0; isv<=svNPoints; isv++)
```
javierrico: Move this 2 [?] spaces to the left

javierrico (changes requested):

javierrico:
Hope this new bunch of comments will need only one more iteration before I can merge (by the way I will have to resolve a couple of conflicts! looking forward to explore new git features... :))

javierrico:
One more thing, please add directory exported_files to the .gitignore file

rcfiles/jointLklDM.rc
dkerszberg: Done.

scripts/jointLklDM.C
dkerszberg: Fixed.

scripts/jointLklDM.C
dkerszberg: Done.

scripts/jointLklDM.C
dkerszberg: Done.

scripts/jointLklDM.C
dkerszberg: Done.

scripts/jointLklDM.C
tjarkmiener: In the Email I sent, I referring to Line 727. Rather than subtracting the inter/extrapolated value at 0, one should subtract the inter/extrapolated at gmin (Minimum of the curve). This comment makes only sense, if our understanding, explained in the Email, is correct.

javierrico: As explained in the e-mail, for gLike it is irrelevant the y-axis offset of the likelihood function. This is a GloryDuck convention and I see no strong reason to change it. Also, it provides an elegant way to know the value of the likelihood function at zero, which is needed for the computation of the detection significance.

scripts/jointLklDM.C
dkerszberg: I had to change that because now I have implemented the "skip DM mass" functionnality that allows to run a rcfile for MAGIC at a mass of 10 GeV without crashing as previously (the goal is to create a glory duck file for all masses, including the one where we are not sensitive). The counterpart is that at 10 GeV nothing will produce signal in MAGIC so we will never reach that point of the code ( "if(imass==0)" ) and for the first mass for which we will reach that point of the code the condition "if(imass==0)" will not be true anymore so the canvas won't be initialize and we won't get any plot (or the program will crash I don't remember now). This change allow everything to work nicely, by just changing this variable from False to True after initialization.

src/GloryDuckTables2019Lkl.cc
dkerszberg: Done.

javierrico (approved):

PR #17, opened 2019-09-17
Prepare gLike for LIV classes.

dkerszberg:


javierrico:
I checked that I can reproduced Segue 1 results with the data in the new format.
Also, I produced the mock data with the new format. Mock data file #1 however contains different values with respect to those of the previous version of the class, because it was produced with a random seed that I don't remember now... Ugly, but no big deal...

PR #18, opened 2019-10-19
Adding new class for line search.

dkerszberg:


include/Iact1dUnbinnedLkl.h
```diff
@@ -69,31 +72,31 @@ class Iact1dUnbinnedLkl : public virtual Lkl
   inline const TH1F*    GetHAeffOff()         const {return fHAeffOff;}
   inline const TH1F*    GetHdNdEpBkg()        const {return fHdNdEpBkg;}
   inline const TH1F*    GetHdNdEpFrg()        const {return fHdNdEpFrg;}
-  inline const TH1F*    GetHdNdEpSignal()     const {return fHdNdEpSignal;}
+  virtual inline TH1F*  GetHdNdEpSignal()     const {return fHdNdEpSignal;}
```
javierrico: All Getters returning pointer should be declared constant, e.g. in this case:
virtual inline const TH1F* GetHdNdEpSignal
any reason why you have removed it? This affects many other Getters in this class and also in the new LineSearch class

include/Iact1dUnbinnedLkl.h
```diff
@@ -69,31 +72,31 @@ class Iact1dUnbinnedLkl : public virtual Lkl
...
   inline const TH1F*    GetHdNdEpSignalOff()  const {return fHdNdEpSignalOff;}
-  inline const TH1F*    GetHdNdESignal()      const {return fHdNdESignal;}
-  inline const TGraph*  GetGEreso()           const {return fGEreso;} 
-  inline const TGraph*  GetGEbias()           const {return fGEbias;}
+  inline TH1F*          GetHdNdESignal()      const {return fHdNdESignal;}
+  virtual inline TGraph*  GetGEreso()         const {return fGEreso;}
```
javierrico: Why have you done this virtual? 
At first I thought that because you needed to redefine it for LineSearch, but that doesn't seem to be the case...
Check other occurrences  of this

include/LineSearchLkl.h
```diff
@@ -0,0 +1,59 @@
...
+ public:
+  
+  // constructors
+  LineSearchLkl(TString inputString="");
+  
+  // destructor
+  virtual ~LineSearchLkl();
+
+  inline  Float_t  GetRelativePeakIntensity() const {return fRelativePeakIntensity;};
+  inline  Float_t  GetBkgRegionWigth()        const {return fBkgRegionWidth;};
```
javierrico: Check typo in the function name "Wight" -> "Width"

include/LineSearchLkl.h
```diff
@@ -0,0 +1,59 @@
...
+
+  Int_t CheckHistograms(Bool_t checkdNdEpBkg=kTRUE);
+  Int_t ResetdNdESignal();
+  Int_t SetdNdESignalFunction(TString function,Float_t p0=0,Float_t p1=0,Float_t p2=0,Float_t p3=0,Float_t p4=0,Float_t p5=0,Float_t p6=0,Float_t p7=0,Float_t p8=0,Float_t p9=0);
+  Int_t SetdNdESignalFunction(TF1* function,Float_t emin=0,Float_t emax=1e9);
+
+ private:  
+
+  TH1F*    fHdNdEpBkg;              //-> dN/dE'dt vs E' for background events (normalized)
+  TH1F*    fHdNdEpSignal;           //-> dN/dE' vs E' histogram for signal events (normalized)
```
javierrico: Why do you need to declare fHdNdEpBkg and fHdNdEpSignal again here? the are already declared in Iact1dUnbinnedLkl. I'm actually surprised this compiles without even a warning...

src/LineSearchLkl.cc
```diff
@@ -0,0 +1,816 @@
+/* ======================================================================== *\
+!
+!   Author(s): Daniel Kerszberg  08/2019 <mailto:dkerszberg@ifae.es>
+!              Tomohiro Inada    08/2019 <mailto:tomohiro@icrr.u-tokyo.ac.jp>
+! 
+\* ======================================================================== */
+
+//////////////////////////////////////////////////////////////////////////////
+//
+// LineSearchLkl
```
javierrico, line 10: Add some documentation please

src/LineSearchLkl.cc
```diff
@@ -0,0 +1,816 @@
...
+//
+// Check that the needed histograms are present and that we are ready 
+// for calling minimization 
+//
+// If needed, convolute dNdESignal*Aeff with Eres and Ebias
+// to get dNdEpSignal
+//
+// Return 0 in case of success, 1 otherwise
+//
+Int_t LineSearchLkl::CheckHistograms(Bool_t checkdNdEpBkg)
```
javierrico: I don't understand why you need this function
The call to  ComputeBkgModelFromOnHisto() can be done in Make checks, and then I think CheckHistograms does not need to be overridden, and the default Iact1dUnbinnedLkl::CheckHistograms will be called. 
Please have a look how this is done in Iact1dBinnedLkl, which is similar in many ways to LineSearchLkl

javierrico (changes requested):
I'm leaving for the weekend, so I stop here in case you can have a look to this bunch of comments.

Issue #23, opened 2020-01-29
Improving IactEventListIrf and Iact1dUnbinnedLkl I/O and interface


cosimoNigro:
The `IactEventListIrf` currently acts as a mere interface between the `gLikeInputs` ROOT files (that are shipped with this repository) and the `Iact1dUnbinnedLkl` analysis class. 
A user cannot produce gLikeInputs himself with the code in this repository. 
More specifically: 

* `IactEventListIrf` misses a `Write()` method to write to disk its output. In PR #24 I am going to show how to define a `IactEventListIrf` object using a `FITS` file I generated with `gammapy`. Once created, I'd like to save it to disk and then to load it in a `Iact1dUnbinnedLkl` object for a likelihood analysis but this is not possible at the moment! 

* Additionally I cannot pass the `IactEventListIrf` instance to the `Iact1dUnbinnedLkl`as no method is given for this. Checking the `Iact1dUnbinnedLkl` setters one can `SetAeff()`, `SetEResoAndBias()`, and `SetObsTime()` but the On and Off event lists cannot be passed.
Indeed in `Iact1dUnbinnedLkl.cc` the IactEventListIrf is directly generated from the `"inputfile="` string (see [see here](https://github.com/javierrico/gLike/blob/master/src/Iact1dUnbinnedLkl.cc#L293)).

Resuming: `Iact1dUnbinnedLkl` reads from disk but `IactEventListIrf` cannot write!
I propose either to add a `Write()` method to `IactEventListIrf`or to add a `SetIactEventListIrf()` method to the `Iact1dUnbinnedLkl`. Or both!

cosimoNigro:
Can I close also this? Now the input for `IactEventListIrf` is separated from `Iact1dUnbinned`

javierrico:
I think so, yes

PR #24, opened 2020-01-29
Initialise IactEventListIrf from FITS files

cosimoNigro:
This PR adds a script to initialise a `IactEventListIrf` (a `gLikeInput` for the MARS users) starting from unbinned spectral dataset in FITS files that I have generated with `gammapy` mimicking gLikeInputs (i.e. lists of On and Off event times and energies, effective area, resolution and bias as a function of the true energy).

These `FITS` files are added in `data/fits`, they are produced reducing with `gammapy` four Crab Nebula observations released in DL3 by H.E.S.S., available [here](https://www.mpi-hd.mpg.de/hfm/HESS/pages/dl3-dr1/) and in [gammapy-extra](https://github.com/gammapy/gammapy-extra/tree/master/datasets/hess-dl3-dr1).

To pass this input to the `Iact1dUnbinnedLkl` we have either to modify `IactEventListIrf` to write to disk or `Iact1dUnbinnedLkl` to have a setter method for an entire `IactEventListIrf` object. See my issue #23.


javierrico:
I'd rather wait for you to include the contents of scripts/testInputFITS.C in the proper class (I think IactEventListIrf according to our off-line discussion. Then there will be no reason to have this macro in the master branch, right?

cosimoNigro:
Yes, in my mind best way to implement it is this:
- introduce a second constructor that does the check of the input file:
```C++
void IactEventListIrf::IactEventListIrf(TString inputFile){
    if (inputFile.EndsWith(".root")){
        // init ON and OFF event lists and IRFs from the ROOT file
        // TFile ....     
        // set fOnSample, fOffSample, fHaeff, fEreso ...
    }
    if (inputFile.EndsWith(".fits")){
        // init ON and OFF event lists and IRFs from the FITS file
        // TFITSHDU ...
        // set fOnSample, fOffSample, fHaeff, fEreso ...
    }
    else{
        // some message about invalid input
    }
}
```
- Then we go into `Iact1dUnbinnedLkl.cc` and replace lines 292-293
```C++
TFile* ifile = new TFile(path+(path==""?"":"/")+inputfileName,"READ");
IactEventListIrf* dataSet = (IactEventListIrf*) ifile->Get("IactEventListIrf");
```
with
```
IactEventListIrf* dataSet = new IactEventListIrf(inputFileName);
```

Alternatively, without adding a second constructor, we can create a `Load` or `Read` method that initialises all the attributes
```C++
void IactEventListIrf::Load(TString inputFile){
    // check the type of input files
    // set fOnSample, fOffSample, fHaeff, fEreso ...
}
```

@javierrico second constructor or `Load()` method, which one do you prefer?

javierrico:
Yes, I like that
But what you propose is so more natural than current 292-293, that it is possible that I tried already that and somehow didn't work, so please try first changing those lines by the proposed ones and test it with the current root files.

javierrico:
About the Load() as you like better, I think there is no practical difference, right?

cosimoNigro:
After discussing with @javierrico we realised that the test ROOT files actually already contain an instance of `IactEventListIrf` stored, not a plain ROOT object. Therefore it is non-obvious how to read that in the constructor of the class itself. This is the reason for the two lines
```C++
TFile* ifile = new TFile(path+(path==""?"":"/")+inputfileName,"READ");
IactEventListIrf* dataSet = (IactEventListIrf*) ifile->Get("IactEventListIrf");
```
in the `Iact1dUnbinnedLkl.cc`.

We should work to establish an input ROOT format and load the data from files containing plain ROOT objects rather than `gLike` classes (I am opening issue #25 on this). 

What I did in the meanwhile is to create a `LoadFITS` method in `IactEventListIrf` and then call it when `Iact1dUnbinnedLkl` interpret the `inputfileName` string, if the latter terminates with `.fits`.

You can do this simple test now
```C++
root [0] Iact1dUnbinnedLkl* unbn = new Iact1dUnbinnedLkl("inputfile=data/fits/obs_id_23523_unbinned_spectrum.fits");
root [1] unbn->PlotHistosAndData()
```
it should work fine and display that the events and IRFs are correct loaded.

src/Iact1dUnbinnedLkl.cc
```diff
@@ -289,66 +289,77 @@ Int_t Iact1dUnbinnedLkl::InterpretInputString(TString inputString)
...
+      fOnSample[i] = TMath::Log10(eventOnE);
+    }
+  for(Int_t i=0;i<fNoff;i++)
+    {
+      dataSet->GetOffEntry(i);
+      fOffSample[i] = TMath::Log10(eventOffE);
     }
   
+  // extract and copy IRFs
+  if(dataSet->GetHAeff())
```
cosimoNigro: @javierrico, for all these histograms you were checking on the number of entries, e.g.
```C++
if(dataSet->GetHAeff()->GetEntries())
```
I removed the `GetEntries` because when I load the FITS files I do not create the empty histograms for the objects I am not going to use (e.g. `fHAeffOff`, `fHdNdEpBkg`). Using `GetEntries()` will cause a crash in this case. It is the same to check on the null pointer directly rather than on the entries of the histogram because each time you would eventually try to use one of this "secondaries" IRFs or counts, you will always check for their existence first.

Makefile
```diff
@@ -5,7 +5,7 @@ INCLDIR    = ./include
 OUTDIR     = ./out
 LIBDIR     = ./lib
 CXX        = g++ -fPIC
-ROOTLIBS   = `root-config --libs` -lMinuit
+ROOTLIBS   = `root-config --libs` -lMinuit -lFitsio
```
javierrico: Does this need that ROOT has been compiled with cfitsio?
what happens if it hasn't?
Is it possible to check and send a warning and an explanation what to do in order to get root compiled correctly?

scripts/testDMSearches.C
```diff
@@ -20,8 +20,8 @@ void testDMSearches()
   const Double_t DlogJ        = 0;     // [GeV^2 cm^-5] statistical error in log_10 of J-factor of the assumed DM source
   const Double_t mass         = 1000.; // [GeV] mass of the DM particle
   const TString  dNdEFileName = TString(Form("./DM/dNdE/Cirelli/dNdESignal_bb_%.1fmass.root",mass)); // dN/dE input file  
-  const TString  inputFile1   = "./data/genericIact_dataIRF_01.root";  // input file with event list and their associated IRF
-  const TString  inputFile2   = "./data/genericIact_dataIRF_02.root";  // input file with event list and their associated IRF
+  const TString  inputFile1   = "./data/root/genericIact_dataIRF_01.root";  // input file with event list and their associated IRF
```
javierrico, line 23: Change the path also in the rc file(s)

src/Iact1dUnbinnedLkl.cc
```diff
@@ -289,66 +289,77 @@ Int_t Iact1dUnbinnedLkl::InterpretInputString(TString inputString)
...
+  IactEventListIrf* dataSet = new IactEventListIrf();
+  if(inputfileName.EndsWith(".root"))
     {
-      cout << "Iact1dUnbinnedLkl::InterpretInputString Warning: no IactEventListIrf object in file " << inputfileName << ", trying the CTAIRF format" << endl;
-      delete ifile; ifile=NULL;
-      if(ReadCTAIRF(path+"/"+inputfileName))
-	return 1;
+      TFile* ifile = new TFile(path+(path==""?"":"/")+inputfileName,"READ");
+      dataSet = (IactEventListIrf*) ifile->Get("IactEventListIrf");
+      if(!dataSet) // check if the ROOT file was correctly interpreted
```
javierrico: Do you know if it is foreseen that CTA IRF's are distributed as DL3-compatible files?
I think it makes a lot of sense, right? then we would not need all this part.
If not, maybe we could also think of a better way to resolve this, now it's very ugly...

src/Iact1dUnbinnedLkl.cc
```diff
@@ -289,66 +289,77 @@ Int_t Iact1dUnbinnedLkl::InterpretInputString(TString inputString)
...
+        {
+          cout << "Iact1dUnbinnedLkl::InterpretInputString Warning: no IactEventListIrf object in file " << inputfileName << ", trying the CTAIRF format" << endl;
+          if(ReadCTAIRF(path+"/"+inputfileName)) 
+            return 1;
+        }
+      ifile->Close();
+      delete ifile;
     }
+  else if(inputfileName.EndsWith(".fits"))
+    {
```
javierrico: According to the optimal [i.e. mine] style, brackets and stuff should not be included for single-line blocks...

src/Iact1dUnbinnedLkl.cc
javierrico: As we have discussed, this check cannot be removed because then empty histograms will be copied into Iact1dUnbinnedLkl class and then the class will try to use them (because it checks if the pointer is or not NULL and if it's not it assumes there are meaningful contents, which would be not the case if you remove the check)

src/IactEventListIrf.cc
```diff
@@ -24,6 +24,12 @@ static const Double_t gEpmin = 1e00; // [GeV] default value of minimum E_est
 static const Double_t gEpmax = 1e06; // [GeV] default value of maximum E_est
 static const Int_t   gBuffSize = 512000; // buffer size for Ntuples
 
+// static functions, helpers to load data from the FITS Header Data Units (HDUs)
```
javierrico, line 37: All this I don't check, I assume it is correct since you are the expert...

javierrico (changes requested):

cosimoNigro:
For the moment this PR is suspended as a patch in ROOT is needed to read a full migration matrix stored in a FITS file (see Issue #25). Before I was storing only resolution and bias but that is not optimal. Once the patch is implemented in ROOT I'll get back to this PR.

cosimoNigro:
The ROOT patch has been approved and the PR completed. 
gLike can now read the FITS data produced by gammapy (and the variable-length arrays stored in the migration matrix HDU columns).
An issue remains related to the ROOT version, see the discussion in Issue #25. 
Once we decide what to do with the ROOT version @javierrico can proceed to review and approve this PR.

javierrico:
Can you please go over the open conversations of this PR and reply to my latest questions/comments? thanks!

cosimoNigro:
Hi @javierrico,

with this latest commits I have isolated with preprocessor directives all the functions using `TFITSHDU` as you requested. 
You can compile the code even if you do not have the latest ROOT release.

The patch in `TFITSHDU`  we use to read FITS files will appear in the [6.21.02](https://github.com/root-project/root/releases/tag/v6-21-02) version which has not yet binaries available (and is not available on conda either).
The latest official release is 6.20.04, the next - including my patch - should be released in the next two weeks.

So, if you want to test the FITS functionalities you have to download one of the recent tags from git and build it yourself.

I checked that scripts and the test macro work correctly with both version 6.20.04 (without FITS patch and functionalities that I installed from conda) and version 6.21.02 (with the FITS patch and functionalities that I built myself).

I added a note in the README.md about the ROOT version and wrote a [page of documentation](https://github.com/javierrico/gLike/wiki/Data-format-for-analysis-of-IACT-data) about the input data formats in the wiki.

You and @dkerszberg might want to check if you can also run the scripts with even older ROOT version.

cosimoNigro:
Ah, I moved the initialisation of `IactEventListIrf` outside of `Iact1dUnbinnedLkl`. 
Before `IactEventListIrf` was initialised within the likelihood definition, opening the input ROOT file:
https://github.com/javierrico/gLike/blob/master/src/Iact1dUnbinnedLkl.cc#L292

Now instead can be more elegantly initialised on its own:
https://github.com/cosimoNigro/gLike/blob/read_fits_inputs/src/Iact1dUnbinnedLkl.cc#L292

I think it makes more sense, and additionally `IactEventListIrf` can be initialised in a more practical way, independent of the Likelihood. For example, opening a ROOT terminal
```C++
root [0] IactEventListIrf *dataset = new IactEventListIrf("dataset1", "root dataset", "data/root/genericIact_dataIRF_01.root")
(IactEventListIrf *) 0x7fd67608c570
root [1] dataset->Print()
Energy Range: 59.2 - 10000 GeV 
ON/OFF Norm (tau): 1.00402 +- 0.00599 (Prob.: -1)
Observation Time: 162000 s 
root [2] dataset->PlotOverview()
```
gLike_root_input


The problem is that I had to drop this `ReadCTAIRF` function.
https://github.com/javierrico/gLike/blob/master/src/Iact1dUnbinnedLkl.cc#L298 
If the ROOT input file is not a "gLike input" (i.e. if it does not contain a saved instance of `IactEventListIrf`) then this function would try to read another ROOT format. 
But to do so it will use other functions of `Iact1dUnbinnedLkl` 
https://github.com/javierrico/gLike/blob/master/src/Iact1dUnbinnedLkl.cc#L1166
that was not easy to cut / copy into `IactEventListIrf`.

It says this function is meant to read `IFAE CTA prod 2 file`. Are these files somewhere, are they still used?

I suggest to remove this function in another PR, or to add CTA data support directly into `IactEventList` from scratch.

dkerszberg:
Hi @cosimoNigro @javierrico 
I managed to compile this gLike version on Ubuntu 18.04 and root v5.34.36. No problem to report!
And I support the idea of implementing the CTA-IRF option (in a different PR), I think now is the right time to do it given all the recent progress towards increased compatibility with various software/format :-)

javierrico:
Hi, 
I am having a problem, not sure if related with your changes or with my root installations or what:
- I can compile the master branch with no problem
- I try to compile your read_fits_inputs branch that I get from gitHub and compilation of source cc files goes right, but after that I get this error:

```
Compiling src/GloryDuckTables2019Lkl.cc...
g++ -fPIC -I./include `root-config --cflags` -c src/GloryDuckTables2019Lkl.cc -o out/GloryDuckTables2019Lkl.o
Generated out/GloryDuckTables2019Lkl.o 

Generating dictionary gLikeDict...
/cern/root/root/bin/rootcint -f out/gLikeDict.cc -c `pwd`/./include/Lkl.h `pwd`/./include/ParabolaLkl.h `pwd`/./include/PoissonLkl.h `pwd`/./include/JointLkl.h `pwd`/./include/Iact1dUnbinnedLkl.h `pwd`/./include/Iact1dBinnedLkl.h `pwd`/./include/IactEventListIrf.h `pwd`/./include/FermiTables2016Lkl.h `pwd`/./include/TemplateLkl.h `pwd`/./include/MIACTEventListIRF.h `pwd`/./include/GloryDuckTables2019Lkl.h `pwd`/./include/gLikeLinkDef.h
Error: Too many '}' tmpfile:7287:
Error: Missing closing brace for the block opened around line 7285.
Error: Unexpected end of file (G__exec_statement()) tmpfile:7288:
Warning: Error occurred during reading source files
Warning: Error occurred during dictionary source generation
!!!Removing out/gLikeDict.cc out/gLikeDict.h !!!
Error: /cern/root/root/bin/rootcint: error loading headers...
make: *** [out/gLikeDict.cc] Error 1
```

My root version is 5.34.36:
```
jrico@Filipo:~/Dropbox/DM/gLike> ls -la $ROOTSYS/..
total 311432
drwxr-xr-x   8 root   wheel   256B  6 ago  2019 ./
drwxr-xr-x   4 root   wheel   128B  2 jun  2015 ../
lrwxr-xr-x   1 root   wheel    14B  6 ago  2019 root@ -> root_v5.34.36/
drwxrwxr-x@ 47 root   wheel   1,5K 14 may  2018 root_v5.34.32/
-rw-r-----@  1 jrico  staff    72M 25 jun  2015 root_v5.34.32.source.tar.gz
drwxrwxr-x@ 47 root   wheel   1,5K 20 jun  2018 root_v5.34.36/
```

Do you know what is the problem?

cosimoNigro:
I think it is your ROOT, can you try with another version?

cosimoNigro:
Sorry I meant the particular ROOT installation that you are using, not the ROOT version.

I get the same error on pic
Screenshot 2020-06-03 at 17 16 07

Screenshot 2020-06-03 at 17 16 48

strangely though, @dkerszberg reports having compiled the `read_fits_inputs` with `root_v5.34.36`on Linux 18.

Which `g++` version are you using @javierrico and @dkerszberg? 
The one on pic is `g++ (GCC) 4.8.5 20150623 (Red Hat 4.8.5-36)` 
It's the only think I can think to.

dkerszberg:
I'm using: gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)

cosimoNigro:
@javierrico can you retry now?
I added the option "-p" to rootcint, in the makefile.
I can compile in pic.
I think the problem was that older rootcint versions have troubles with two level of preprocessor directives
https://root.cern.ch/root/roottalk/roottalk97/0736.html

I hope these incoveniences make us all reflect on the cost of maintaining bronze-age root versions due to "some" gLike users.

javierrico:
> @javierrico can you retry now?
> I added the option "-p" to rootcint, in the makefile.
> I can compile in pic.
> I think the problem was that older rootcint versions have troubles with two level of preprocessor directives
> https://root.cern.ch/root/roottalk/roottalk97/0736.html

yes! it worked! you're a genius!

> 
> I hope these incoveniences make us all reflect on the cost of maintaining bronze-age root versions due to "some" gLike users.

:rofl:

README.md
```diff
@@ -15,6 +15,13 @@ Follows a non-exhaustive list of examples where gLike is useful (in order of inc
  - Estimating the energy scale of quantum gravity by combining observations of fast gamma-ray flares observed by different ground-based gamma-ray telescopes.
  - ...
  
+### Prerequisites
+`gLike`'s sole dependency is [ROOT](https://root.cern.ch). 
+
+#### Note on IACT analyses with gLike 
+To input [IACT data in FITS format](https://github.com/javierrico/gLike/wiki/Data-format-for-analysis-of-IACT-data#fits-format) `gLike` uses sepcial `TFITSHDU` methods available only in very recent ROOT versions (> 6.20.04).
```
javierrico: Please fix typo in "sepcial"

include/IactEventListIrf.h
```diff
@@ -36,17 +39,27 @@ class IactEventListIrf : public TNamed
   static const Double_t gDefHadVal    ;    //! default value when hadronness is not provided
 
   // constructor
-  IactEventListIrf(TString name="IactEventListIrf",TString title="");
-  
+  void _initialize_me();      // initialize empty resources, used by both constructors
```
javierrico: What is wrong with calling this function "Initialize" or "InitializeMe" and keep style?

include/IactEventListIrf.h
```diff
@@ -36,17 +39,27 @@ class IactEventListIrf : public TNamed
...
+  IactEventListIrf(TString name="IactEventListIrf", TString title="");
+  IactEventListIrf(TString fileName, TString name, TString title);
+
   // destructor
   virtual ~IactEventListIrf();
 
+  #if ROOT_VERSION_CODE > ROOT_VERSION(6,20,04)
```
javierrico: In principle I would like it better if the function is declared in any case (i.e. no preprocessor directive needed here in the .h file) and then when you define the function at the .cc file, you decide what to do, namely: you read the file normally for ROOT versions high enough or, e.g. issue an error otherwise. I think moving the error to the execution time is better than having it at the compilation time, because you can try to handle it and look for alternatives. 
If you have a different opinion I would like to know...

javierrico (changes requested):
Please fix this couple of things and I will merge
And please reply to the comments! even with an "ok"

README.md
cosimoNigro: Corrected!

include/IactEventListIrf.h
cosimoNigro: Sorry, I saw it in ROOT and adopted it.
Renamed `_initialize_me() -> Initialize()`

include/IactEventListIrf.h
cosimoNigro: Ok, the preprocessor directives are now only in `IactEventListIrf.cc` and `IactEventList.h` has the `LoadFITSFile` function declared.

cosimoNigro:
I also have added a check in the Makefile that will add the `-lFitsio` option to `ROOTLIBS` when compiling with the most recent ROOT versions
https://github.com/cosimoNigro/gLike/blob/read_fits_inputs/Makefile#L11

This way you do not have to add it manually if you want to use the `TFITS` class and the makefile will not complain if you have an old ROOT compiled without it.
Thanks a lot @dkerszberg for helping with this!

javierrico (approved):

PR #29, opened 2020-02-03
Brane tension

TjarkMiener:
Dear @javierrico 

I want to take advantage :-P of your recent GitHub activity and add some branon related changes to gLike. As shown in [Fig.7; right panel](https://arxiv.org/pdf/1905.11154.pdf), you can translate the sgimav limits to the brane tension parameter space. For now, the limits for the brane tension will be calculated and printed to the console. Let me know, if you think that gLike should generate a f(m_{DM}) plot by default.

Cheers,
Tjark

### DIFF
--- a/scripts/computeCLBands.C
+++ b/scripts/computeCLBands.C
@@ -33,6 +33,7 @@
 #include <iostream>
 #include <iomanip>
 #include <fstream>
+#include <sstream>
 #include <vector>
 
 #include "TPRegexp.h"
@@ -40,6 +41,7 @@
 #include "TH1.h"
 #include "TF1.h"
 #include "TString.h"
+#include "TObjString.h"
 #include "TCanvas.h"
 #include "TLatex.h"
 #include "TFile.h"
@@ -337,6 +339,7 @@ void computeCLBands(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int
       Double_t braneTensionVal[nmass];
       for(Int_t imass=shift;imass<nmass;imass++)
         {
+          Float_t mdm = massval[imass];
           // release the memory of channelval and brval
           delete [] channelval;
           delete [] brval;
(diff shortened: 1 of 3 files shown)


scripts/jointLklDM.C
```diff
@@ -910,6 +915,13 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
   for(Int_t imass=0;imass<nmass;imass++)
     cout << svSenVal[imass] << (imass<nmass-1? "," : "");
   cout << "};" << endl;
+  if(!channel.CompareTo("branon",TString::kIgnoreCase))
+    {
+      cout << "Double_t braneTensionLimit[nmass]  = {";
+      for(Int_t imass=0;imass<nmass;imass++)
+        cout << 0.001*TMath::Power((braneTensionVal[imass]*1.167*TMath::Power(10., -17.))/svLimVal[imass], 1./8.) << (imass<nmass-1? "," : "");
```
dkerszberg: Here I would suggest to use 1.1664 instead of 1.167 following the actual precision to which this constant is known (http://pdg.lbl.gov/2019/reviews/rpp2018-rev-phys-constants.pdf).

dkerszberg: Also I think it would be good to quote in a comment where the numbers comes from and what it represents.

dkerszberg: Also explains from which formula does the 0.001 (GeV to TeV), the 10^(-17) and the 1/8 comes from (just pointing to a reference where the formula is displayed is enough I think)

scripts/computeCLBands.C
```diff
@@ -288,6 +331,52 @@ void computeCLBands(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int
...
+          nChannels = 9;
+          channelval = new TString[nChannels];
+          brval = new Double_t[nChannels];
+          // call the function that computes the branching ratios and save them in channelval and brval
+          compute_branonBR(mdm,nChannels,channelval,brval,braneTensionVal[imass]);
+        }
+
+      cout << "Double_t bt0sigma_"+channel+"[nmass]  = {";
+      for(Int_t imass=shift;imass<nmass;imass++)
+        cout << 0.001*TMath::Power((braneTensionVal[imass]*1.167*TMath::Power(10., -17.))/sv0sigma[imass], 1./8.) << (imass<nmass-1? "," : "");
```
dkerszberg: Also add in this section, add the same comment about the origin of the different values.

scripts/jointLklDM.C
tjarkmiener: Done!

TjarkMiener:
Ready to test and merge @dkerszberg :)

scripts/computeCLBands.C
```diff
@@ -288,6 +331,74 @@ void computeCLBands(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int
...
+          nChannels = 9;
+          channelval = new TString[nChannels];
+          brval = new Double_t[nChannels];
+          // call the function that computes the branching ratios and save them in channelval and brval
+          compute_branonBR(mdm,nChannels,channelval,brval,braneTensionVal[imass]);
+        }
+
+      // To translate the sigmav values from [cm^3 s^-1] to [GeV^-2], the sigmav limits have to be divided by
+      // h_bar^2 c^3 = (6.58*10^-25 [GeV s])^2 * (3.0*10^10 [cm s^-1])^2 = 1.17*10^-17 [GeV^2 cm^3 s^-1].
+      // The 0.001 factor is translating the brane tension limit from GeV to TeV.
```
dkerszberg: To take advantage of all the newly fixed constant :-) may I suggest something like:

// To translate sigmav values in [cm^3 s^-1] into brane tension values in [GeV^2], the following multiplication factor has to be applied
// h_bar^2 c^3 with dimension [GeV^2 s^2] * [cm^3 s^-3] = [GeV^2 cm^3 s^-1] with
// h = 6.626 07015×10^−34 J s ; e = 1.602176634×10^−19 C;  c = 299792458x10^2 cm s^−1
// following the values provided by the Particle Data Group in
// http://pdg.lbl.gov/2019/reviews/rpp2018-rev-phys-constants.pdf
// The final conversion formula also contains:
// - a factor (0.001) to translate the brane tension limit from GeV to TeV
// - a power 1/8 to relate sigma to f (see equation 7 of https://arxiv.org/abs/hep-ph/0302041)

Then you can compute once for all precisely this factor by doing:
factor_name = ((h/(2. * TMath:Pi())) / (e*10^9)) * (299792458x10^2)^3
and use it everywhere it's needed in the following lines

If you agree with this comment, it should be applied everywhere in the code where it is used.

scripts/jointLklDM.C
```diff
@@ -910,6 +915,29 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
   for(Int_t imass=0;imass<nmass;imass++)
     cout << svSenVal[imass] << (imass<nmass-1? "," : "");
   cout << "};" << endl;
+
+  // To translate the sigmav values from [cm^3 s^-1] to [GeV^-2], the sigmav limits have to be divided by
+  // h_bar^2 c^3 = (6.58*10^-25 [GeV s])^2 * (3.0*10^10 [cm s^-1])^2 = 1.17*10^-17 [GeV^2 cm^3 s^-1].
+  // The 0.001 factor is translating the brane tension limit from GeV to TeV.
```
dkerszberg: Potentially same comment as above.

scripts/jointLklDM.C
```diff
@@ -980,6 +1029,37 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
...
+  // canvas for brane tension plots
+  TCanvas* branoncanvas = NULL;
+  if(!channel.CompareTo("branon",TString::kIgnoreCase))
+    {
+      branoncanvas  = new TCanvas("branoncanvas","Brane tension limits",800,800);
+
+      TH1I *dummybtlim = new TH1I("dummybtlim",Form("f ULs vs mass"),1,massval[0],massval[nmass-1]);
+      dummybtlim->SetStats(0);
+      dummybtlim->SetMinimum(0.001*TMath::Power((braneTensionVal[0]*1.17*TMath::Power(10., -17.))/plotmax, 1./8.));
+      dummybtlim->SetMaximum(0.001*TMath::Power((braneTensionVal[nmass-1]*1.17*TMath::Power(10., -17.))/plotmin, 1./8.));
```
dkerszberg: Potentially same comment as above.

scripts/computeCLBands.C
tjarkmiener: The sigmav values are in  [cm^3 s^-1] and with dividing h_bar^2 c^3 we are getting the **sigmav** values in [GeV^-2]. The brane tension has units of [GeV].

Having a variable for the defined/pre-computed "exact" constant is a good point. Since the Particle Data Group provided us with h_bar = 6.582119569x10^-22 [MeV s] = 6.58211956910x^-25 [GeV s] and c = 299792458.0 [m s^−1] = 299792458.0x10^2 [cm s^−1] we can just use these values and pre-compute h_bar^2 c^3 = 1.167329990x10^-17 [GeV^2 cm^3 s^-1]. I think there is no need to start the calculation from h in [J s]. Just to keep it as simple as possible.

dkerszberg:
Thanks a lot @TjarkMiener! For me this PR is ready to be merged @javierrico.

javierrico (approved):
Nothing else to add

PR #30, opened 2020-02-04
added break for the case that lklval is 1e99 in line 456

mgaug:


javierrico:
Hi @mgaug 
thanks for the commit
I am having a hard time understanding in which situation the original code lines produce an endless loop and why this fixes it, can you please describe briefly? Thanks!

mgaug:
Hi Javi,

The problem seems to be related with return value of "CallMinimization" in line 675, for the case that the fit does not converge, and the counter reaches "maxcounts". In that case, CallMinimization returns 9e99, which is, on the other hand, greater than "fgmin+4*fErrorDef" on line 451. 

This leads to an infinite loop in line 451, although function "CallMinimization" is exempt of infinite loops, because of the limitation in calls to the minuit through maxcounts.

Note that the next loop, in line 461, does not suffer from this problem, because it requires the return argument of CallMinimization to be *less* than "fgmin+fErrorDef", which is the case if "lklval" becomes 1e99. 

BTW, why is the contraction of the lower end limited to "fgmin+4*fErrorDef", and the upper end only to fgmin+(1)*fErrorDef ?

javierrico:
Thanks Markus
Sorry for being silent so long time. I am trying to catch up with pending work now. 

After having a careful look to the code I still cannot understand how there can be an infinite loop in that while-loop: the value of g for which CallMinimization is called is different at each loop, getting asymptotically closer to the value of g producing the minimum likelihood, for which the minimization should converge (it has for sure converged at least once earlier in the code execution, otherwise it would not be the minimum lkl value).

It would be extremely helpful for me if you could send me the rc file that produces the infinite loop, or at least the log file where you can see it. I hope it's not too far in the past so that you cannot find it anymore. Thanks!

javierrico:
This problem has now become an issue, some more investigation is needed before we solve it 
Closing the PR without merging

Issue #27, opened 2020-02-03
We lack a basic testbed for new developments


javierrico:
@cosimoNigro, as we have discussed offline, could you think and eventually implement a testbed consisting in running a defined macro with defined rc file so that the result can be contrasted with the output plots that we would store somewhere in the gitHub repository?

cosimoNigro:
Ok, I'll make of this my priority rather than PR #24.
In this way, once we have defined a minimum unit of testing, we can see what happens if we change the functions for the input.

javierrico:
Ok, good

PR #31, opened 2020-02-05
Added test suites for PoissonLkl.h, JointLkl.h and IactEventListIrf.h

cosimoNigro:
Hello @javierrico, 

In Issue #27 you expressed the wish to have some test unit.

This PR introduce a `test` directory containing (at the moment) some scripts to test the `PoissonLkl`, the `JointLkl` and the `IactEventListIrf` classes.

Taking a look at [the testing tutorial of google](https://github.com/google/googletest/blob/master/googletest/docs/primer.md), testing consists in checking if one established result can be re-produced once change in the code have been implemented (see the [factorial example](https://github.com/google/googletest/blob/master/googletest/docs/primer.md#simple-tests)).
One can also check if class functionalities (especially those creating new objects) behave as expected.

Following Google's nomenclature I call `suite` a collection of functions performing the test relative to a given class, so `TestSuitePoissonLkl.cc` groups all the functions performing tests on `PoissonLkl.h` and so on and so forth. I don't think it is worth to create a class for testing, I think we can be fine with collections of functions for the time being.

Let us assume we want to see if future changes in the code change the results from the current version `v00.08.00` - that we assume to be stable.    
In `TestSuitePoissonLkl.cc` I hardcode some of the results of a case of usage of `PoissonLkl` with the current version. I let the test re-run a `PoissonLkl` minimisation with the same input and then compare its results with the expected (hardcoded) one.
Same goes for `TestSuiteJointLkl.cc`, where I instead use two `PoissonLkl` objects. 

If in the future we change the code of the likelihood classes and such a change modifies the results of the minimisation, then the test will crash. Unless it's a bug of course, in that case we will update the hardcoded results.

In `TestSuiteIactEventListIrf.cc` instead I do some minimal testing on the `IactEventListIrf` class. I check that all the members exposed through getters (ON sample, OFF sample, effective area etc..) are not null pointers. Additionally, for the file `data/genericIact_dataIRF_01.root`, I check that the number of entries of each of these members are the same that I can read with this version of the code.
Again, if future changes of the class will leave one of the members initialised, or will impede the correct loading of one of these members, the test will fail.
We can eventually check also for the IRF values or for some random content of the ON / OFF event list. I leave that to further discussion, and to another day.

All the available tests are exposed in the `test/main.C` macro. If you want to run it from my fork please run it in debug mode, i.e.
```bash
root[0] .x test/main.C++g
```
only in this mode the `assert` will effectively crash the execution (otherwise it will continue unaffected even if `assert(false)` is met).

Let me know what you think.
Also @all the other users are welcome to suggests more tests to perform.

P.S. 
I formatted the code with clang, with the `llvm` style, using 2 spaces for the indentation and no tabs
```bash
clang-format -style="{BasedOnStyle: llvm, IndentWidth: 2, UseTab: Never}" -i test/*
```
see if you like it, I think your "optimal" style is not available and you have to cope with this fact.


javierrico:
I am merging this pull request, just a few comments:
- please make sure that the asserting is done in a way that makes 100% clear to the developer making the check that it has failed and where, independently on how she runs the check (compiled, not compiled, with debugging mode, etc...)
- include in the test one that runs jointLklDM.C for few selected masses/channels/processes (decay/annihilation) on one of the mock data files and check the results at the level of: i) the LklVsG function ii) the limits to <sv> or decay time
- Produce documentation for: i) the way the check needs to be performed by anyone before asking for a PR; ii) the way developers introducing new features should also provide their own checks

PR #36, opened 2020-06-10
Change hard-coded very-high likelihood values for static constants

javierrico:
Hi @cosimoNigro and @dkerszberg,
I am reviewing PR #30 and I will probably use this alternative PR to implement the necessary changes to solve the issue pointed out there.
As a first step, I have changed all places where very high numbers appeared hard-coded in the code (like, e.g. returning 9e99 as lkl value when there was no convergence) and replace them by two static constants, check Lkl.h
I think it's a safe change, but could you please have a look and also think if for your classes this could make any difference?
Thanks


dkerszberg:
Hi @javierrico, it shouldn't make any difference, as far as I can tell I will just need to adopt this new convention in the Line Search class. One question: 1e80 (instead of 1e99 previously) for very high likelihood values is just to be able to more easily distinguish the very high case from the overflow case?

Do you want us to approve this PR as it is or do you want to continue to work on it before accepting it?

javierrico:
> Hi @javierrico, it shouldn't make any difference, as far as I can tell I will just need to adopt this new convention in the Line Search class. One question: 1e80 (instead of 1e99 previously) for very high likelihood values is just to be able to more easily distinguish the very high case from the overflow case?

Yes, that is the reason

> 
> Do you want us to approve this PR as it is or do you want to continue to work on it before accepting it?

Approving means merging? I still need to do some more work before merging

dkerszberg:
Yes I meant merging. Then I/we wait :-)
And thanks for the explanation, as said above all good for me so far with the changes.

dkerszberg:
I recently encountered problems that this PR is trying to fix, indeed running gLike on many samples these days I noticed of few occurrence of this message:

`Lkl::FindGLowAndGUpp (JointLkl_00) Message: -2logL(gupp) = 1.9e+100, gupp lowered from -inf to -inf (fLklMin = 197.978)`

repeating endlessly (I even let one run for 5h and a few 10^5 occurrences just to be sure) for obvious reasons.

I think it is the symmetric of the problem encountered for glow (see #30 ). At least for me this problem at high energy, with low statistics and is actually happening for masses for which tau doesn't converge to a positive value but to "-inf" (very pathological indeed).

I will test to limit tau to positive values and if it improves the situation I will push the modification.

javierrico:
Hi @dkerszberg, can you please send me the rc file (and maybe mc seed if this is for simulations) to reproduce this effect? thanks!

PR #38, opened 2020-08-27
Shortcut for the GloryDuck files in simulation mode

TjarkMiener:
This PR will add the seed to the input string for the GloryDuck txt files.

### DIFF
--- a/exec/computeCLBands.cc
+++ b/exec/computeCLBands.cc
@@ -195,7 +195,7 @@ int main(int argc, char* argv[])
   for(Int_t ifile=1;ifile<nSimuFiles+1;ifile++)
     {
       // file name
-      TString seedtag  = Form("_%05d",ifile);
+      TString seedtag  = Form("_%d",ifile);
       TString filename = simusPath+label+"_MC_limits"+seedtag+".root";
       cout << "Opening file " << filename << endl;
       Int_t exist = gSystem->Exec("ls " + filename + " > /dev/null");
--- a/exec/jointLklDM.cc
+++ b/exec/jointLklDM.cc
@@ -381,6 +381,13 @@ int main(int argc,char* argv[])
 	}
       else if(classType.CompareTo("GloryDuckTables2019Lkl")==0)
 	{
+          if(isSimulation)
+            {
+              TPMERegexp re(".txt");
+              UInt_t nfields = re.Split(inputString);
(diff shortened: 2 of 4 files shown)


cosimoNigro:
Hi @TjarkMiener,
sorry, I received a notification only now about this.
The idea would be that the script `jointLklDM.C` becomes deprecated in the next release and we start using the executable `jointLklDM`, whose source is in `exec/jointLklDM.cc`.

Have you tried to build gLike with cmake as I requested [some time ago here](https://github.com/javierrico/gLike/pull/33#issuecomment-654131474), to check if the new executable works as the old script? 
To me it worked, as far as I could check with the mock data in this repository.

Maybe we should implement these fixes and changes already in `exec/jointLklDM.cc` rather than in the old `.C` that we'll discard anyhow?

dkerszberg:
Hi @cosimoNigro 
I also missed that I have to admit. Can we keep both for a few more weeks? Tjark and I are finalizing the GloryDuck analysis and I would like to run it the "old way" first (all my scripts are ready for that) but I would also be happy to run it the "new way" (with the executable) to cross-check that they work both exactly the same :-)
@TjarkMiener could you push the same changes in the exec/jointLklDM:cc file?

Apart from that missing bit (thanks Cosimo!) the PR is ok for me!

cosimoNigro:
Ok for me if you change both, thank you!

TjarkMiener:
Hi @cosimoNigro! Sorry, I missed that too! I will have a closer look after the GloryDuck analysis and come back to you.

javierrico:
Ok, I see just one problem, having to do with the release structure rather than the code:
If you want to use a code with these changes to produce publishable results, the changes should go into a release that can be properly referenced.
Now, I don't want to start a whole new _release branch_ (check [this](https://github.com/javierrico/gLike/wiki/Branch-releases-structure-and-log) if you don't know what that is) just for this minor thing, so I suggest that you request the pull to the latest release branch, i.e. v00.09, can you do that @TjarkMiener ?

javierrico:
Please don't leave for too late moving this PR to the v00.09 branch
There is another minor bug fix recently merged there and  would like to issue the new release including both PRs, and also merging all changes to the master branch. Thanks

TjarkMiener:
Is it solved by changing the base branch from master to v00.09? Sorry for opening #43.

PR #39, opened 2020-09-01
Fix bug when reading a file

dkerszberg:
I think the path to the file was deleted by mistake when the fits PR was done. My usual analysis were not running anymore after I updated to the latest master branch, this bugfix made it work again. Can you check it @cosimoNigro ?

### DIFF
--- a/src/Iact1dUnbinnedLkl.cc
+++ b/src/Iact1dUnbinnedLkl.cc
@@ -290,7 +290,7 @@ Int_t Iact1dUnbinnedLkl::InterpretInputString(TString inputString)
     }
   
   // open and read input files with data and IRFs
-  IactEventListIrf* dataSet = new IactEventListIrf("dataSet", "", inputfileName);
+  IactEventListIrf* dataSet = new IactEventListIrf("dataSet", "", path+(path==""?"":"/")+inputfileName);
 
   // extract info from file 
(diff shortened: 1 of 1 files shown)


javierrico:
Can you request this to be pulled in branch v00.09 please?
Read [here](https://github.com/javierrico/gLike/wiki/Branch-releases-structure-and-log) why

dkerszberg:
Sure no problem.
When I open a PR for this from my master branch it also includes 4 commits on the README file (concerning zenodo) done by @javierrico and @cosimoNigro . So just to be 100% sure you only want this commit to be included in v00.09 release and not the one regarding zenodo right?

javierrico:
In general, you should make your bug fix in your v00.09 branch and ask the PR from there, then the zenodo commits will not appear.
But in this case I think we want also the changes in the README (particularly the doi) to be included in all releases, so please include also those in your PR

dkerszberg:
> In general, you should make your bug fix in your v00.09 branch and ask the PR from there, then the zenodo commits will not appear.

Yes this is what I had in mind.

> But in this case I think we want also the changes in the README (particularly the doi) to be included in all releases, so please include also those in your PR

And this is what I wanted to check with you :-) Thanks for the clarification, the new PR is opened!

PR #48, opened 2021-02-09
added the neutrino channels from Cirelli

TjarkMiener:
Ref: JCAP 1103:051,2011 (arXiv:1012.4515)

http://www.marcocirelli.net/PPPC4DMID.html


javierrico:
Hi Tjark, thanks for taking care. I have a few questions.

First is about nomenclature: can you clarify what "Ve" and "nue" mean?
I guess that nue represents the nu_e fluxes for the DM DM -> nu_e nu_e reaction, right?
But this does not work for "Ve", and also, what is V?

Second is, how did you generate the files?

Thanks

TjarkMiener:
hi Javi,
the nomenclature was taking from the Cirelli table. I generated the files with the macros you shared with me some time ago. Macro names are cirelliSpectraToIncludeFile and cirelliSpectraToRootHistos.
Correct, nue is what you wrote above and needed for the neutrino project. 
For the secluded DM part, you can find a nice study here: [Profumo et al.](https://arxiv.org/pdf/1711.03133.pdf). V would be the mediator, which will then decay into SM particles. I was more thrilled about testing this with MAGIC one month ago, so we could drop it and not include it in the repo. I could test it locally though (if time allows).

I see, I forgot to update the scripts to parse correctly the channels. I will do so soon!

javierrico:
Ok about the macros
To make sure, these are dNdE for gamma rays, right? and NOT for neutrinos (which would be the final state)
About nomenclature, it is no extrictly the same as for the already existing channels. You probably you want to use:
"nuenue", "numunumu", "nutaunutau" to comply with previous nomenclature
As for the mediator case, I suggest: "VV-4e", "VV-4mu", "VV-4tau"
There are many empty files, because no gammas are produced, those cases should be excluded, we don't want empty files.
Actually, do you know by which process gammas are generated in the nunu cases? I really have no idea...

TjarkMiener:
Thanks for the comments! I will have a further look about the empty files and that the macro is taking the proper dNdE. The nomenclature looks good to me! For my defense ;-), I adopted the nomenclature from the Cirelli table.

About the gammas produced in the annihilation into nunu, I found this explanation in [Cirelli et al. page 17](https://arxiv.org/pdf/1012.4515.pdf). It also worth to check Fig. 4 on page 16.
> One can see that the portion of energy which goes into gamma rays and e± is often the most important one and always dominates over the energy fraction of the hadronic final states for all the channels. This is especially relevant in the context of extragalactic gamma rays signatures, where the energy fraction in e± is quickly converted to gamma rays due to Inverse Compton radiation.For the channels involving μ+μ− and τ+τ− and of course for the ν ̄ν channels, the portion of energy carried away by neutrinos becomes the dominant one.  The fractions are rather independent of the mass of the DM particle,  with some exceptions.  For example, in the νν channels,  primary neutrinos start to radiate gamma rays and charged leptons due to radiative weak corrections when M_DM is above the electroweak scale (i.e.  for the outer pie in the figure) and this increases the energy fraction of γ and e±.

javierrico:
- Thank you for the explanations about gamma production in neutrino annihilation channels, it is clear now

- I hope you mean that MY suggested nomenclature looks ok to you... This will make it more homogeneous with the already existing channels. Consider that that nomenclature was defined before Cirelli even existed...

- The empty files should be empty, because no gamma-rays are produced for those channels and masses, but then do not commit them into gLike. Instead, you need to define the corresponding thresholds for the different channels at jointLklDM.cc (lines 216-229) and maybe corresponding lines at the macro version. Can you also please take care?

TjarkMiener:
Hi @javierrico 
I included the requested changes. I didn't added the thresholds to exec/computeCLbands because it wasn't even with scripts/computeCLBands.

javierrico:
Hi @TjarkMiener 
I still see a lot of empty files, at least some of which should not be (e.g.  DM/dNdE/Cirelli/dNdESignal_VV-4e_100.0mass.root) and actually were not in previous commits
Can you please double-check?

TjarkMiener:
How are you checking if a file is empty or not? My (not sophisticated) approach was to open some files with the TBrowser. And then detecting empty files from the file size after verifying some examples.

For this particular example (DM/dNdE/Cirelli/dNdESignal_VV-4e_100.0mass.root), the TBrowser is displaying a dNdE. Is this file empty on your machine?
Screenshot 2021-03-17 at 14 06 48

javierrico:
Go to "files changed" on top of this page and you will see how for some it says they are empty
Captura de pantalla 2021-03-17 a las 14 22 27

TjarkMiener:
Ah i see. But github says that all DM/dNdE/Cirelli/*.root files are empty. This is very strange, because you can see here (and my previous comment) that those files are actually not empty:
Screenshot 2021-03-17 at 14 30 25

One very strange thing occur, when I checked the file uploads some time ago with the normal SM channels:
Screenshot 2021-03-17 at 14 34 04
Here, github displays the file size properly.

javierrico:
> Ah i see. But github says that all DM/dNdE/Cirelli/*.root files are empty. 

That's really weird, it wasn't happening in the previous commit....
And that was the reason I discovered the fact that the threshold should be implemented
I love when you solve a problem following a completely unrelated clue but for very good reasons, very Chacal like...
Anyway, I'll merge

PR #50, opened 2021-03-09
Adding a factor for the quarks

TjarkMiener:
Hi @javierrico 
Unfortunately, there was a factor of 3 missing for the quarks. Viviana and me noticed it when comparing the BRs to the cta branon paper. I will rerun the branon analysis with this fix. I won't expect a noticeable difference, since the bosons are dominating in our energy range anyways. Could you please release a new version with this fix, so we can tag this gLike version in the paper? Sorry for the inconvenient!


javierrico:
Hi Tjark, sorry, since this is a fix that you need a release for, I need you to make this pull request to branch v00.09, thanks

PR #57, opened 2021-05-24
Change of convention for Glory Duck and various small fixes

dkerszberg:


dkerszberg:
This PR improves convergence and stability with respect to the problems described in https://github.com/javierrico/gLike/pull/36#issuecomment-735875203
But this PR doesn't supersede the above PR #36 (it uses the old convention with hard-coded values).

.gitignore
```diff
@@ -16,5 +16,4 @@ build/
 out/
 lib/
 plots/
-results/
```
javierrico, line 19: Private change right? no need to push it

Makefile
```diff
@@ -46,7 +46,7 @@ $(OUTDIR)/%.o: $(SRCDIR)/%.cc $(HEADERS)
 .PHONY : clean
 clean:
 	@rm -f $(OUTDIR)/*.o
-	@rm -f $(OUTDIR)/$(TARGET)Dict.*
```
javierrico, line 49: Makefile will be deprecate in the next release, so I'd skip this too

rcfiles/jointLklDM.rc
```diff
@@ -9,7 +9,7 @@ I/O CONFIGURATION
 #################
 
 # Label to indentify this analysis (for naming plots and output files)
-jointLklDM.Label: GenericIACT_Annihilation_bb_Unbinned_GisPositive_JisFixed
+jointLklDM.Label: GenericIACT_Annihilation_bb_Unbinned_and_Binned_GisPositive_JisFixed
```
javierrico, line 12: I'm rather removing reference to binned or unbinned

rcfiles/jointLklDM.rc
```diff
@@ -33,7 +33,7 @@ jointLklDM.exportSvLogStep: TRUE
 jointLklDM.exportDataPath: exported_files
 
 # Path (will be added to all specified files and directories in this rcfile)
-jointLklDM.path: ./
+jointLklDM.path: .
```
javierrico, line 36: Safer the original way...

scripts/jointLklDM.C
```diff
@@ -873,7 +873,7 @@ void jointLklDM(TString configFileName="$GLIKESYS/rcfiles/jointLklDM.rc",Int_t s
       //////////////////////////////////////////////////////////////
       if (exportData)
         for (Int_t isv=0; isv<=svNPoints; isv++)
-          vlkl2D[isv][imass] = grLklParabola[imass]->Eval(svScan[isv]) - grLklParabola[imass]->Eval(0);
+          vlkl2D[isv][imass] = grLklParabola[imass]->Eval(svScan[isv]);
```
javierrico, line 876: Ok, you know better than me, adopted, but in the .cc file, since the macro will be deprecated in the next release

src/Lkl.cc
```diff
@@ -560,7 +565,10 @@ Double_t Lkl::MinimizeLkl(Double_t g,Bool_t gIsFixed,Bool_t isVerbose,Bool_t for
   SetMinuitLink();
 
   // assign value to g and fix it if requested
-  fMinuit->DefineParameter(gGParIndex,fParName[gGParIndex],g+fGShift,fParDelta[gGParIndex],0,0);
+  if(fParName[gGParIndex]=="tau")
```
javierrico, line 568: This is not a good thing to do... you are thinking on a  very precise definition of tau with a natural range of uncertainties below 10, but that is by no means compulsory
Why do you want to do this? maybe we can find alternative ways to do it...

src/Lkl.cc
```diff
@@ -585,6 +593,7 @@ Double_t Lkl::MinimizeLkl(Double_t g,Bool_t gIsFixed,Bool_t isVerbose,Bool_t for
   // call minimization (with requested verbosity)
   Int_t strategy = (!gIsFixed)*2; // when it's fixed, we can relax the strategy since there is no interest in errors
   Double_t lkl = CallMinimization(g,isVerbose,strategy);
+  if(lkl>8e99) return -1000.;
```
javierrico, line 596: why??

src/Lkl.cc
```diff
@@ -637,7 +646,7 @@ Double_t Lkl::CallMinimization(Double_t g,Bool_t isVerbose,Int_t strategy)
   arglist[0] = 10000;
   iflag = -1;
   Int_t counter = 0;
-  const Int_t maxcounts = 50;
+  const Int_t maxcounts = 300;
```
javierrico, line 649: where did you see this was necessary?

src/Lkl.cc
```diff
@@ -453,14 +453,17 @@ void Lkl::FindGLowAndGUpp(Double_t& glow,Double_t& gupp,Bool_t centerAtZero)
       Double_t save  = glow;
       glow+=TMath::Abs(glow-gmin)/2.;
       cout << "Lkl::FindGLowAndGUpp (" << GetName() << ") Message: -2logL(glow) = " << lklval << ", glow rised from " << save << " to " << glow <<" (fLklMin = " << fgmin << ")"<< endl;
+      if(lklval >= 1e99) break;
```
javierrico, line 456: I don't see how this breaking the loop in a value of g with no lkl value is a good thing...
Maybe by continuing the loop you will eventually reach the region with good lkl values. In principle this should happen, since at each step you get closer to gmin
This should converge unless really gmin is **exactly** the minimum value of g with well defined likelihood value...

src/Lkl.cc
```diff
@@ -453,14 +453,17 @@ void Lkl::FindGLowAndGUpp(Double_t& glow,Double_t& gupp,Bool_t centerAtZero)
...
     }
   
   // expand the lower end
   while((lklval=MinimizeLkl(glow,kTRUE,kFALSE,kTRUE))<fgmin+fErrorDef || (centerAtZero && glow>0))
     {
-      Double_t save  = glow;      
+      Double_t save  = glow;
+      if(fGIsPositive && glow<0) break;
```
javierrico, line 463: Time saver, adopted

src/Lkl.cc
```diff
@@ -483,6 +486,7 @@ void Lkl::FindGLowAndGUpp(Double_t& glow,Double_t& gupp,Bool_t centerAtZero)
       Double_t save  = gupp;      
       gupp-=TMath::Abs(gupp-gmin)/2.;
       cout << "Lkl::FindGLowAndGUpp (" << GetName() << ") Message: -2logL(gupp) = " << lklval << ", gupp lowered from " << save << " to " << gupp <<" (fLklMin = " << fgmin << ")"<< endl;
+      if(lklval >= 1e99) break;
```
javierrico, line 489: This can never happen because of the loop condition, right?

src/Lkl.cc
```diff
@@ -491,6 +495,7 @@ void Lkl::FindGLowAndGUpp(Double_t& glow,Double_t& gupp,Bool_t centerAtZero)
       Double_t save  = gupp;      
       gupp+=(gupp<0? 2*TMath::Abs(gupp) : TMath::Abs(gupp));
       cout << "Lkl::FindGLowAndGUpp (" << GetName() << ") Message: -2logL(gupp) = " << lklval << ", gupp rised from " << save << " to " << gupp <<" (fLklMin = " << fgmin << ")"<< endl;
+      if(lklval >= 1e99) break;
```
javierrico, line 498: As in the previous comment, this should never happen, right?

src/Lkl.cc
```diff
@@ -661,16 +670,23 @@ Double_t Lkl::CallMinimization(Double_t g,Bool_t isVerbose,Int_t strategy)
 
       // try changing precision
       if(iflag!=0 || TMath::IsNaN(GetParErr(gGParIndex)))
-	fMinuit->DefineParameter(gGParIndex,fParName[gGParIndex],g,fParDelta[gGParIndex]/10.*TMath::Power(1.8,counter), 0, 0);
```
javierrico, line 664: Same comment as above, related to the special treatment for all parameters called "tau"

src/Lkl.cc
```diff
@@ -661,16 +670,23 @@ Double_t Lkl::CallMinimization(Double_t g,Bool_t isVerbose,Int_t strategy)
...
+        {
+          if(fParName[gGParIndex]=="tau")
+            fMinuit->DefineParameter(gGParIndex,fParName[gGParIndex],g,fParDelta[gGParIndex]/10.*TMath::Power(1.8,counter), 0, 10);
+          else
+            fMinuit->DefineParameter(gGParIndex,fParName[gGParIndex],g,fParDelta[gGParIndex]/10.*TMath::Power(1.8,counter), 0, 0);
+        }
       
       counter++;
-      if(counter==maxcounts)
+      if(iflag != 0 && counter==maxcounts)
```
javierrico, line 681: Taken

src/Lkl.cc
```diff
@@ -661,16 +670,23 @@ Double_t Lkl::CallMinimization(Double_t g,Bool_t isVerbose,Int_t strategy)
...
 	cout << "Lkl::CallMinimization (" << GetName() << ") Warning: No convergence reached for g = " << g << " after " << maxcounts << " trials: check your -2logL curves for features" << endl;
+
     }
 
   delete [] arglist;
 
-  return (counter<maxcounts? GetLklVal() : 9e99);
+  if(iflag != 0) return 9e99;
+  else return (counter<=maxcounts? GetLklVal() : 9e99);
```
javierrico, line 689: Taken adapted version

src/Lkl.cc
```diff
@@ -840,7 +856,12 @@ void Lkl::SetParameters(const Char_t** parname, Double_t* parstart, Double_t* pa
       fParErr[ipar]     = 0;
       fIsParFixed[ipar] = kFALSE;
       if(fMinuit)
-	fMinuit->DefineParameter(ipar, fParName[ipar], fParStart[ipar], fParDelta[ipar], 0, 0);
```
javierrico, line 843: Same comment as above, related to the special treatment for all parameters called "tau"

javierrico (commented):
Hi, sorry for taking so long to reply to you
I have included some of your suggested changes in the branch alternative_PR30, which I want to merge soon to the master branch. Some other suggestions I do not understand. I have commented your suggested changes in those senses, please have a look in case you want to counter-reply...

.gitignore
dkerszberg: If I'm not mistaken the pushing of "results" was a private change as gLike don't create such output folder by default, hence my suggestion to remove it

Makefile
dkerszberg: Can't we still fix it (I think it's another "mistake" change from a previous PR)? At least it would be right in the latest version where it exists :-) (as of now the "make clean" removes something that doesn't exist). Anyhow I'm ok to deprecate it soon.

rcfiles/jointLklDM.rc
dkerszberg: Ok!

rcfiles/jointLklDM.rc
dkerszberg: If you want, but it's not necessary as the macro/exec takes care properly of that and as of now it makes some ugly "//" in the output

scripts/jointLklDM.C
dkerszberg: It's just the recent (at least few months) change of convention to not bound the lkl curve to zero at <sigmav>=0 but at the minimum of the parabola. Same as above, shouldn't we have it also in the macro so the last version of it is in agreement with the adopted convention?

src/Lkl.cc
dkerszberg: I saw some "tau=inf" sometimes hence this ugly trick, I agree to discard it for now. Will try to find an example of this problem with the latest fixes, i think the line paper will give me a good opportunity, I'll keep you posted soon (working on it now).

src/Lkl.cc
dkerszberg: Glory Duck sometimes need more than 150 iterations and I think a few times >200 was needed. I'm unsure why though... but the convergence is healthy, it just take a lot of iterations sometimes. Maybe in the future we can define a more clever maxcounts (based on TMinuit feedback?)?

src/Lkl.cc
dkerszberg: It's linked to line 251: if (TMath::Abs(Lkl::MinimizeLkl() + 1000.) < 1.e-6) return 0;
This was a trick we introduced some years ago to allow gLike to exit properly in this case (otherwise it's crashing in the next step). We said we will change it for something better the day where we have a proper error handling system...
Btw line 560 also has "return -1000.;" (just in case you decide to tackle this now :-) ).

src/Lkl.cc
dkerszberg: "Maybe by continuing the loop you will eventually reach the region with good lkl values." --> this is what I empirically observed, the "break" works well in practice although I agree should not be needed in principle...
" This should converge unless really gmin is exactly the minimum value of g with well defined likelihood value..." --> I don't understand this, let's discuss it orally :-)

src/Lkl.cc
dkerszberg: Sometimes I have good tricks up my sleeves :-D

src/Lkl.cc
dkerszberg: What in the loop condition prevents it?

src/Lkl.cc
dkerszberg: I'm almost certain it does, I will try to find a past example of this.

src/Lkl.cc
dkerszberg: Agreed!

src/Lkl.cc
dkerszberg: Making a lot of testing I found some very edgy cases :-D the converged status but still failed really annoyed me when it meant loosing 24h ^^

src/Lkl.cc
dkerszberg: Ok!

src/Lkl.cc
dkerszberg: Agreed!

dkerszberg:
@javierrico can you push your commits on alternative_PR30 so I can see them in #36 ? As of now they are only in your computer I think.

.gitignore
javierrico: Ah yes, I thought you were _adding_ the directory. 
But you are right, this should be removed, also the following ones now that the source and building directories are clearly separated thanks to the cmake building procedore:
-plots
-results
-**/dNdEpSignal
-exported_files
-**/*.pcm
-old
-.ipynb_checkpoints

Makefile
javierrico: ok

rcfiles/jointLklDM.rc
javierrico: I did not know about that feature of jointLklDM, but still, I think it is better to add the final "/" so that it is clear at first sight that we are referring to a directory. I do not care that much about double or triple "/"

scripts/jointLklDM.C
javierrico: Ok

src/Lkl.cc
javierrico: Ok, changed, for the time being, the -1000 to the new gLklValOverflow value

src/Lkl.cc
javierrico: Ok

src/Lkl.cc
javierrico: Please send me a configuration for reproducing such error, thanks!

src/Lkl.cc
javierrico: No, I mean like before, that this should not generate an infinite loop. I believe you when you say it does, but I'd like to understand how, again I'm gonna need the example...

src/Lkl.cc
javierrico: Same as before

javierrico:
I have included all the changes that we agreed upon in the master branch already
There are two pending issues, for which I will open the corresponding issues, namely:

1. (issue #61): In Lkl::FindGLowAndGUpp, it seems that the while loops some times get infinite. It has been suggested by Markus Gaug in PR #30 and by @dkerszberg  in PR #36 to add a break to each of them when the likelihood value gets very high. @javierrico would like to understand why we are getting those infinite loops and requests @dkerszberg to provide him with a configuration to reproduced the bug
2. (issue # 62): Sometimes, the best-fit value of the Iact1dUnbinned class gets inf. @dkerszberg had suggested (PR #36) to limit its value, but that is not good practice and for sure cannot be done in Lkl.cc as proposed, if anything in Iact1dUnbinnedLkl, for which this parameter is defined. We need to understand why this is happening and find a better solution

Then, if you agree, we can close this PR for the time being and keep working on the pending issues

PR #59, opened 2021-11-04
Add the possibility of fast simulations containing a finite signal component

javierrico:
I have added the possibility of including a signal in the fast simulated samples for Iact1dUnbinnedLkl and Iact1dBinnedLkl classes.
I have also modified exec/jointLklDM.cc [BUT NOT THE CORRESPONDING MACRO!!!] and rcfiles/jointLkl.rc  to use this new feature.
The input parameter (set in the rc-file) is the assumed value of \<sv\>, and to convert it to the expected number of signal events and their distribution in E', the DM fUnitsOfG and dNdEpSignal histogram must have been previously defined. 
Therefore, the previous scheme in which the simulation was produced immediately after the input files were read would not work (except for \<sv\>=0), because fUnitsOfG depends also on the considered mass and annihilation/decay channel. 
Now, when \<sv\> \> 0, an independent simulation has to be produced per mass. Note that although \<sv\> is always the same, the DM particle mass changes and it's always exactly equal to the mass searched for. 
A modification is pending by which we can decouple these two things: the true DM mass used for the generation of events, and the value assumed when building the likelihood.  But this should be enough for producing the plots for replying to the referee of the all-dSph MAGIC paper
@dkerszberg: can you please make a few checks? I suggest the following:
- Simulate a signal with \<sv\> above the sensitivity in a mass range, and check that the best-fit \<sv\> and upper limits you obtain in such cases agree with the simulated value, for binned and unbinned cases
- Check the output plots for consistency with the previous point too. In particular, for sufficiently high value of simulated \<sv\> you should see the signal shape in the On event energy distribution
- For a given mass, use DlogJ = 0 and gIsPositive=FALSE to simulate the case \<sv\> = 0 and see what  is the value of the 95% CL **sensitivity** that you obtain. Then make some 300 simulations of \<sv\>=sensitivity and plot the obtained distribution of best-fit values of \<sv\>. Check that about 5% of the obtained cases are below 0.
- Once you do that you can try some typical DlogJ value and compare what happens

### DIFF
--- a/exec/jointLklDM.cc
+++ b/exec/jointLklDM.cc
@@ -272,7 +272,7 @@ int main(int argc,char* argv[])
   cout << " GeV" << endl;
   
   cout << "*** IRF/DATA PLOTS      : " << (showSamplePlots?   "YES" : "NO") << endl;
-  cout << "*** PARABOlA PLOTS      : " << (showParabolaPlots? "YES" : "NO") << endl;
+  cout << "*** PARABOLA PLOTS      : " << (showParabolaPlots? "YES" : "NO") << endl;
   cout << "*** Plot y-axis range   : " << plotmin << " to " << plotmax << Form(" %s", (isDecay? "s" : "cm^3/s")) << endl;
   cout << "*** EXPORT DATA         : " << (exportData? "YES" : "NO") << endl;
   if(exportData)
@@ -907,7 +907,7 @@ int main(int argc,char* argv[])
   gSystem->Exec(Form("mkdir -p %s/pdf",realPlotDir.Data()));
   gSystem->Exec(Form("mkdir -p %s/jpg",realPlotDir.Data()));
   
-  TString seedTag  = (seed<0? "" : Form("_%d",seed));
+  TString seedTag  = (seed<0? "" : Form("_%04d",seed));
   
   limcanvas->Print(realPlotDir+"root/"+label+"_"+simulationlabel+"_limits"+seedTag+".root");
   limcanvas->Print(realPlotDir+"pdf/" +label+"_"+simulationlabel+"_limits"+seedTag+".pdf");
--- a/exec/jointLklDM.cc
+++ b/exec/jointLklDM.cc
@@ -154,7 +154,7 @@ int main(int argc,char* argv[])
   TString  fPlotsDir         = fInputDataPath+"/"+env->GetValue("jointLklDM.plotsDir","")+"/";
   TString  provval           = env->GetValue("jointLklDM.dNdEpSignalDir","-");
   TString  fdNdEpSignalDir   = fInputDataPath+"/"+provval+"/";
-  Float_t  mcG               = env->GetValue("jointLklDM.mcG",0.);  //assumed value of <sv> for simulations
+  Float_t  mcsv              = env->GetValue("jointLklDM.mcsv",0.);  //assumed value of <sv> for simulations
   TString  massList          = env->GetValue("jointLklDM.MassList","");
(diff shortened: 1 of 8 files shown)


javierrico:
Ah, @dkerszberg I forgot:
please have a look to line 395 of exec/jointLklDM.cc, where there is a condition for simulations in the case of GloryDuckTables2019Lkl class, which I really don't know what it means, but it could be affected by the introduced changes (hopefully not...)

javierrico:
These commits should not change the behavior of the code. They represent a stable reorganization of the code in the way towards being able to generate a signal using a pdf different from the one assumed in the likelihood function

exec/jointLklDM.cc
```diff
@@ -252,7 +257,11 @@ int main(int argc,char* argv[])
   cout << " ** Channel check            : " << normchannel << endl;
   cout << "*** DATA/MC                  : " << simulationlabel << endl;
   if(isSimulation)
-    cout << " ** Seed                     : " << seed << endl;
+    {
+      cout << " ** Seed                     : " << seed << endl;
+      cout << " ** <sv>_MC                  : " << mcsv << " cm^3/s" << endl;
+      cout << " ** m_MC                     : " << mcmass << " GeV" << endl;
```
dkerszberg, line 263: Also print mcChannal and mcProcess?

exec/jointLklDM.cc
```diff
@@ -714,17 +624,29 @@ int main(int argc,char* argv[])
...
+		fullLkl->SetDMDecayUnitsForG(mass);
 	      else	      
-	        fullLkl->SetDMAnnihilationUnitsForG(mass);
+		fullLkl->SetDMAnnihilationUnitsForG(mass);
+	      
+	      // in case simulations were requested with DM cont, do them now
+	      if(isSimulation && mcsv>0 && mcmass>0)
+		{		  
+		  cout << "  * Simulating observations for <sv> = " << mcsv << " and m_DM = " << mcmass << endl;
+
```
dkerszberg: For my understanding, this is where we will need to build the histo for mcmass, mcChannel and mcProcess and pass it on SImulateDataSamples, isn't?

dkerszberg:
> Ah, @dkerszberg I forgot: please have a look to line 395 of exec/jointLklDM.cc, where there is a condition for simulations in the case of GloryDuckTables2019Lkl class, which I really don't know what it means, but it could be affected by the introduced changes (hopefully not...)

It is not affected :-) This piece of code by @TjarkMiener does the nice trick of converting GloryDuck files names from bb_Segue1_MAGIC.txt to bb_Segue1_MAGIC_X.txt where X is the seed number (between 1 and 300 for us) so that we can run the simulations with the same rcfile and don't need to create 300 rcfiles.

exec/jointLklDM.cc
dkerszberg: Also print mcChannel and mcProcess?

dkerszberg:
Hi @javierrico I have also added the same code to the macro as in the executable.

javierrico:
> Hi @javierrico I have also added the same code to the macro as in the executable.

Hi, there are again many changes in the executable, maybe you can commit another version of the macro

javierrico:
With these new commits, the pdf used for the MC signal generation and the one for building the likelihood function are decoupled. Now the default workflow in jointLklDM.cc is to have one true signal simulated with fixed values of mass, <sv> (or 1/tau), channel and process (all to be specified at the rc file) and then do the usual search assuming a given process, channel, and scanning over mass to find limits to <sv> (or 1/tau)

javierrico:
Also, a bug in the display of the histograms in the plot function (I think appearing only for root 6) is solved

PR #76, opened 2024-01-31
Add HDMSpectra

StFroese:
Add root files for [HDMSpectra](https://github.com/nickrodd/HDMSpectra/tree/master)

Lower limit for mass is 500 GeV


TjarkMiener:
Thanks @StFroese! Please correct the file names: 1. there is a f after mass value and 2. there is always a 2 at the end of the filename.

StFroese:
> Thanks @StFroese! Please correct the file names: 1. there is a f after mass value and 2. there is always a 2 at the end of the filename.

Done :shipit:

